PR #65, opened 2023-03-21
build: Use universal build for mac os

juliushaertl:
Make sure to build a universal app bundle on Mac OS so that it has a native binary available when running on apple silicon chips.

Tested with an M1 and works fine though I cannot test intel compatibility of the resulting bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
 		"start": "electron-forge start",
 		"package": "electron-forge package",
 		"package:linux": "electron-forge package --platform=linux",
-		"package:mac": "electron-forge package --platform=darwin",
+		"package:mac": "electron-forge package --platform=darwin --arch=universal",
 		"package:windows": "electron-forge package --platform=win32",
 		"package:all": "electron-forge package --platform=all",
 		"make": "electron-forge make --skip-package",


Ivansss:
I do and it works fine here

package.json
```diff
@@ -19,7 +19,7 @@
 		"start": "electron-forge start",
 		"package": "electron-forge package",
 		"package:linux": "electron-forge package --platform=linux",
-		"package:mac": "electron-forge package --platform=darwin",
+		"package:mac": "electron-forge package --platform=darwin --arch=universal",
```
ShGKme, line 22: If I'm not mistaken, packaging to `Darwin --arch=universal` is possible only on MacOS host. While packaging 2 separated `Darwin-x64` and `Darwin-arm64` is possible on any host...

From Mac users perspective and without publishing to MAS, is universal package better than separated?

juliushaertl: I'd say a universal binary makes it easier for non-technical users, otherwise they'd need to know which binary to download. Other than that separate binaries are of course smaller each, but that would be the only differences I could think of.

juliushaertl: Right, just checked and universal bundle building indeed only works directly on macOS. Let me close this PR then, thought it would be a quick fix, but probably something you have on your list anyways depending on how releases are going to be built in the future.

Issue #53, opened 2023-03-21
No app icon on Linux


ShGKme:
The icon should also be set as the window icon: https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions

l4b4r4b4b4:
Yep can confirm!

System: Ubuntu 22.10
Nextcloud Talk App: v0.2.2

PR #67, opened 2023-03-21
Add window icon for Linux app icon

ShGKme:
Not tested on Linux yet

Fix: #53

### DIFF
--- a/src/app/windowIcon.js
+++ b/src/app/windowIcon.js
@@ -19,47 +19,26 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { BrowserWindow } = require('electron')
-const { BASE_TITLE } = require('../constants.js')
-const { windowIcon } = require('../app/windowIcon.js')
+const path = require('node:path')
+const {
+	isLinux,
+	isWindows,
+	isMac,
+} = require('../shared/os.utils.js')
 
-/**
- *
- * @param {import('electron').BrowserWindow} parentWindow - main window (parent)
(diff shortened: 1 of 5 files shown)


nickvergessen:
Doesn't show on the binary nor the dock :(

src/app/windowIcon.js
```diff
@@ -0,0 +1,44 @@
...
+
+if (isWindows()) {
+	windowIcon = path.join(__dirname, '../img/icons/icon.ico')
+} else if (isMac()) {
+	windowIcon = path.join(__dirname, '../img/icons/icon.icns')
+} else if (isLinux()) {
+	windowIcon = path.join(__dirname, '../img/icon.png')
+} else {
+	// Also use PNG as default
+	windowIcon = path.join(__dirname, '../img/icon.png')
```
Antreesy, line 39:
```suggestion
	windowIcon = path.join(__dirname, '../img/icons/icon.png')
```

According to [docs](https://www.electronforge.io/guides/create-and-add-icons#configuring-installer-icons), this also should be added to forge.config.js:
```js
      name: '@electron-forge/maker-deb', //same for zip, i guess
        config: {
          options: {
            icon: './img/icons/icon.png'
          }
      }
```

But it still doesn't appear on Linux Dock

ShGKme: > According to docs, this also should be added to forge.config.js:

Yes, but we don't plan to build distributable before final release

> //same for zip, i guess

No, zip is just an archive, it doesn't have any icons or OS specific "package" settings.

Issue #70, opened 2023-03-22
User ID shown instead of user display name while in a call


nickvergessen:
**Others** in the call see your UID instead of your display name in the call view on top of the video:
![uid-vs-displayname](https://user-images.githubusercontent.com/213943/226824769-3746dc43-9256-46ae-824a-37dad14aa948.png)

nickvergessen:
Same for temporary messages:
![image](https://user-images.githubusercontent.com/213943/226854653-b3c44960-5b34-4993-afb2-d9a47f671d30.png)

So seems like the initialisation of the actor store is not working as expected:
https://github.com/nextcloud/spreed/blob/3f47c4e9cda4fc9733efb438ceba4ba56a063562/src/store/messagesStore.js#L604-L606

nickvergessen:
Data comes from:
https://github.com/nextcloud/spreed/blob/c09bfbfb509f672249be270fd07f7265ac0506d5/src/App.vue#L426
```js
import { getCurrentUser } from '@nextcloud/auth'
```

PR #72, opened 2023-03-22
Fix wrong display-name property from provisioning getUser response

ShGKme:
### ☑️ Resolves

* Issue #70

In provisioning API `getUser` and `getCurrentUser` methods returns display name with a different property name.
In this PR `@nextcloud/auth/getCurrentUser` tries both properties considering the current name as "legacy".

Related: https://github.com/nextcloud/server/pull/36665

### 🚧 Tasks

- [x] Use legacy `'display-name'` property as a fallback for display name


nickvergessen (approved):
Fixes the issue

Issue #75, opened 2023-03-22
Brute force protection triggered since using the desktop client


juliushaertl:
I have more often slow requests that seem to come from brute force protection since I started using the desktop client. Though I'm not fully sure this is a but in the client, I wanted to create this ticket to start collecting some insights.

Nextcloud logs shows a lot of entries for my IP:

```
{"reqId":"CHIKQ0vqOCzOoMiXX7BJ","level":2,"time":"2023-03-22T17:13:24+00:00","remoteAddr":"77.2.30.xxx","user":"--","app":"core","method":"GET","url":"/index.php/apps/notify_push/uid","message":"Login failed: 'undefined' (Remote IP: '77.2.30.xxx')","userAgent":"--","version":"26.0.0.10","data":{"app":"core"}}
```

I couldn't see the request being issued by the talk desktop client, but the user agent seems to be empty, so might still be the case.

At least the undefined user id would match with what I see in the desktop client network log for the notify_push server web socket:

<img width="649" alt="Screenshot 2023-03-22 at 21 04 07" src="https://user-images.githubusercontent.com/3404133/227026054-f4cd711c-5c4b-4339-89e1-337212740880.png">

ShGKme:
> but the user agent seems to be empty

It not supposed to be empty... UserAgent header is sent.

juliushaertl:
Maybe the request is actually issued by notify_push as the url is also not the same

PR #76, opened 2023-03-22
Fix loosing authentication on notify_push on any ws error

ShGKme:
### ☑️ Resolves

* Issue #75

### 🚧 Tasks

- [x] Add missed credentials on reconnect after error in notify_push client. It caused infinity unsuccessful reconnections
- [x] Refactor notify_push a bit

### DIFF
--- a/src/patchers/nextcloud-notify_push.js
+++ b/src/patchers/nextcloud-notify_push.js
@@ -26,6 +26,7 @@
 
 import { getCapabilities } from '@nextcloud/capabilities'
 import { subscribe } from '@nextcloud/event-bus'
+import axios from '@nextcloud/axios'
 
 /**
  * Get the list of supported notification types as reported by the server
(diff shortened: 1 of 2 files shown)


src/patchers/nextcloud-notify_push.js
```diff
@@ -141,7 +162,7 @@ async function setupSocket({ user, password } = {}) {
 
 		setTimeout(() => {
 			if (window._notify_push_online) {
-				setupSocket()
+				setupSocket({ credentials })
 			}
 		}, 1000 * window._notify_push_error_count)
```
ShGKme, line 167: There credentials were missed

ShGKme:
Although it seems like this problem cause the issue, it is not clear to me, what was the reason for the first WS error started infinity errors loop.

juliushaertl (approved):
Works for me, thanks 💙

Issue #74, opened 2023-03-22
Can not zoom


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open the window on 4k
2. Try to zoom it like 120% or alike for improved readability

### Actual behaviour
There is no option for it, neither does Ctrl + Scroll nor Ctrl + `+` work.
Maybe it's okay and instead a setting ( #27 ) should be implemented, but that should then also be usable with known shortcuts

SystemKeeper:
![image](https://user-images.githubusercontent.com/1580193/226957116-3b7e8d55-0e2c-4cf7-8c52-9ee457928731.png)

Works on Mac?

nickvergessen:
Right, works when opening the menu with `Alt`

ShGKme:
> Ctrl + `+` work.

Did you try Ctrl + SHIFT + `+` ? Without `SHIFT` it works as `CTRL + =`.

nickvergessen:
That's totally unexpected for me. Is it configurable which we could adjust?

PR #77, opened 2023-03-22
Add CommandOrControl + Plus hotkey for zoomIn without SHIFT

ShGKme:
### ☑️ Resolves

* Issue #74

### 🚧 Tasks

By default an Electron application has "CommandOrControl + Plus" hotkey for zooming in. It is not very  intuitive, that without SHIFT pressing "CommandOrControl + Plus" actually works as "CommandOrControl + =". 

- [x] Add additional invisible zoomIn menu item with "CommandOrControl + =" hotkey

Adding invisible menu item to add hotkey alias may look wierd. But it seems to be a common practice to solve this problem in Electron applicatoin in a simple and reliable way.

Also, it was possible to change the hotkey in the existed menu item, but it would change displayed hotkey. This solution keeps displaying original hotkey in menu while both hotkeys works exactly the same.

image

### DIFF
--- a/src/app/app.menu.js
+++ b/src/app/app.menu.js
@@ -93,6 +93,13 @@ function setupMenu() {
 			{ type: 'separator' },
 			{ role: 'resetZoom' },
 			{ role: 'zoomIn' },
+			// By default zoomIn works by "CommandOrControl + +" ("CommandOrControl + SHIFT + =")
+			// Hidden menu item adds zoomIn without SHIFT
+			{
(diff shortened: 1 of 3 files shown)


nickvergessen:
`Ctrl` + Scroll is also addable in this way? Then we would be state-of-the-art again with all other programs

nickvergessen (approved):
Approving as it's already making it much more discoverable

nickvergessen (approved):

Issue #102, opened 2023-03-28
The About dialog can't be closed


e-caste:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. click on top right user icon
2. click on "About"
3. try to close the dialog

### Expected behaviour
The dialog closes when clicking outside it on the shadowed part of the window or on the X in one of the top corners (I can't see it so maybe it should be added).

### Actual behaviour
Can't close the dialog.

## Desktop client

**Talk Desktop client version:** v0.2.1

**Operating system:** macOS

**Operating system version:** Ventura 13.3

**Microphone available:** yes

**Camera available:** yes

## Server

**Nextcloud version:** (see status page: `/status.php`) 26.0.0

**Talk app version:** (see apps admin page: `/index.php/settings/apps`) 16.0.1

**Custom Signaling server configured:** yes/no and version (see additional admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`) no

**Custom TURN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#turn_server`) yes

**Custom STUN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#stun_server`) yes

### Client log

<details>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

e-caste:
@ShGKme 
> Do you mean the window opens without window controls?  

Exactly, see the screenshot below. I've tried clicking everywhere but the only way to close the About window is to quit the app and re-open it.

<img width="1276" alt="Screenshot 2023-03-28 at 16 19 29" src="https://user-images.githubusercontent.com/48513706/228268818-e8353b29-bb68-4841-ad76-2988a6209edd.png">

ShGKme:
@e-caste Thanks for the bug report, fixed

PR #103, opened 2023-03-28
Fix unclosable About window on Mac - make About window closable by Escape and a new close button

ShGKme:
### ☑️ Resolves

* Issue #102

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

By default and by design native modal windows on MacOS doesn't have controls and not closable.

- [x] Add close button to the about window
- [x] Close About window by Escape


Ivansss (approved):
Tested and works 👍

Issue #84, opened 2023-03-27
Notification if an update is available


jakobroehrl:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

nickvergessen:
Definitely planned for the long run, but won't happen for the first couple of previews.

ShGKme:
Closed by #106

PR #106, opened 2023-03-29
New release notification

ShGKme:
### ☑️ Resolves

* Issue #84 

### 💁 Description

Every 2h checks if there is a new release on GitHub releases and notify with a push notification. Click on the notification opens the latest release page.

### 🖼️ Screenshots

image

### DIFF
--- a/src/app/githubReleaseNotification.service.js
+++ b/src/app/githubReleaseNotification.service.js
@@ -1,0 +1,127 @@
+/*
+ * @copyright Copyright (c) 2023 Grigorii Shartsev <grigorii.shartsev@nextcloud.com>
+ *
+ * @author Grigorii Shartsev <grigorii.shartsev@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


nickvergessen (approved):
I modified it to check the freshly created release org repo

src/app/githubReleaseNotification.service.js
```diff
@@ -0,0 +1,127 @@
...
+		clearInterval(schedulerIntervalId)
+	}
+}
+
+module.exports = {
+	getLatestReleaseVersion,
+	notifyAboutNewVersion,
+	checkForNewVersion,
+	setupReleaseNotificationScheduler,
+	stopReleaseNotificationScheduler,
```
nickvergessen, line 126: grafik


PS most of them seem to not be used outside, do you export by default? Or does it make sense to not export it?

ShGKme:
@nickvergessen Before merging, do you think ~2h interval is ok?

nickvergessen:
yes, would have even opted for "in the morning only"

ShGKme:
> yes, would have even opted for "in the morning only"

User settings are planned for later this week.

PR #107, opened 2023-03-29
fix(l10n): Add fake info.xml so the translationtool can convert back …

nickvergessen:
…from *.po files to json and js

https://github.com/nextcloud/docker-ci/blob/33beb897e4152895109f1657272d5346701863a2/translations/translationtool/src/translationtool.php#L381-L395

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0"?>
+<info>
+	<id>talk_desktop</id>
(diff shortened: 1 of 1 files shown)


ShGKme (approved):

ShGKme:
> …from *.po files to json and js

If it is configurable, we need only json files, not js

nickvergessen:
Unluckily not at the moment

Issue #73, opened 2023-03-22
Application blackout after click on the conversation


AndyXheli:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open Nextcloud Talk.exe
2. Select Conversation
3. Application dose black

### Expected behaviour
Should load the conversation

### Actual behaviour

https://user-images.githubusercontent.com/59488153/226942952-08469fdb-7c08-4d60-89c8-13823bdf32f1.mp4

## Desktop client

**Talk Desktop client version:** v1.0.0
0.2.1
**Operating system:** Windows / MacOS / Ubuntu
Windows
**Operating system version:** Windows 11 / (In case of Linux, provide the distribution.)
Widnows 11
**Microphone available:** yes/no
Yes
**Camera available:** yes/no
Yes
## Server

**Nextcloud version:** (see status page: `/status.php`)
Nextcloud 26
**Talk app version:** (see apps admin page: `/index.php/settings/apps`)
16.0.1
**Custom Signaling server configured:** yes/no and version (see additional admin settings: `/index.php/index.php/settings/admin/talk#signaling_server`)

**Custom TURN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#turn_server`)

**Custom STUN server configured:** yes/no (see additional admin settings: `/index.php/settings/admin/talk#stun_server`)

### Client log

<details>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

ShGKme:
I cannot reproduce on Windows 11

Could you please add javascript console log?

1. Open devtools with CTRL+SHIFT+S or from Menu/Window/Toggle Dev Tools
2. Console Tab
3. Clean with CTRL+L
4. Open a conversation

AndyXheli:
Sorry so i open Nextcloud Talk desktop then i do the following 

Open devtools with CTRL+SHIFT+S or from Menu/Window/Toggle Dev Tools
Console Tab
Clean with CTRL+L
Open a conversation

AndyXheli:
[...]

PR #141, opened 2023-04-13
Fix session loosing caused by different `set-cookie` casing

ShGKme:
### ☑️ Resolves

* Issue #73

### 💁‍♂️ Description

Session is defined in cookies. Server may return cookies in different casing e.g. `set-cookie`, `Set-Cookie`, `SET_COOKIE`. Previously only `set-cookie` was checked (I used to think, Electron force all headers to lower case...).

In this PR interceptor finds all `/set-cookie/i` headers.

### 🖼️ Screenshots

No changes

### 🚧 Tasks

- [x] Check cookies in headers with case ignore.

### DIFF
--- a/src/app/webRequestInterceptor.js
+++ b/src/app/webRequestInterceptor.js
@@ -63,12 +63,20 @@ function enableWebRequestInterceptor(serverUrl, {
 	 */
 	function persistCookies(details) {
 		// OPTIONS will have new session - ignore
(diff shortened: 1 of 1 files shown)


ShGKme:
Thanks @AndyXheli for providing server for reproducing the bug!

nickvergessen:
All headers are case insensitive btw

nickvergessen (approved):

PR #142, opened 2023-04-14
fix(patchers): fix importing images and sounds using `@nextcloud/router`

ShGKme:
### ☑️ Resolves

* Issue: https://github.com/nextcloud/spreed/pull/9276#discussion_r1166415172

### 🖼️ Screenshots

No visual changes in the current version

### 🚧 Tasks

Previously all the files were requested from the server, except sounds - they were imported as a local file from a copy. This is a problem because:
- Talk images should also be imported from the local files, not requested
- Talk sounds should not be copied to this repo - we can require them from the `spreed` repo during the build. Only the `notifications` app's sounds could be copied.

- [x] Update `generateFilePath` patch from `@nextcloud/router`:
  - [x] Always require images and sounds for `'spreed'` app from the `spreed` repository
  - [x] Require sounds only for the `'notifications'` app from the local copy
  - [x] Remove unused copy of the `spreed`'s sounds
- [x] Patch `imagePath` from `@nextcloud/router` to use patched `generateFilePath`


src/patchers/nextcloud-router.js
```diff
@@ -52,13 +50,42 @@ export function generateRemoteUrl(service) {
...
+			return require(`@talk/img/${filename}.png`)
+		} else if (ext === '.jpg') {
+			return require(`@talk/img/${filename}.jpg`)
+		} else if (ext === '.jpeg') {
+			return require(`@talk/img/${filename}.jpeg`)
+		} else if (ext === '.webp') {
+			return require(`@talk/img/${filename}.webp`)
+		} else if (ext === '.ogg') {
+			return require(`@talk/img/${filename}.ogg`)
+		}
```
Antreesy, line 75:
```suggestion
		switch (ext) {
		case '.svg':
			return require(`@talk/img/${filename}.svg`)
		case '.png':
			return require(`@talk/img/${filename}.png`)
		case '.jpg':
			return require(`@talk/img/${filename}.jpg`)
		case '.jpeg':
			return require(`@talk/img/${filename}.jpeg`)
		case '.webp':
			return require(`@talk/img/${filename}.webp`)
		case '.ogg':
			return require(`@talk/img/${filename}.ogg`)
		}
```
I would like to go with switch-case here

Antreesy (approved):
Tested at mentioned PR.
Works properly!

src/patchers/nextcloud-router.js
nickvergessen:
```suggestion
		if (ext === '.svg' || ext === '.png' || ext === '.jpg'  || ext === '.jpeg' || ext === '.webp' || ext === '.ogg') {
			return require(`@talk/img/${filename}.${ext}`)
		}
```
Or am I missing something?

nickvergessen: Why are we even limiting it to a set of extensions?

ShGKme: I prefer to avoid the `switch-case` statements. It is close to `goto label:`:
- You can forget `break` or `return` and break everything
- You cannot make a more complex condition
While it isn't making the code shorter or faster.

If we want to make the code shorter and more like a mapping from the extension to a requiring here, I may propose mapping with object:

```js
const requiresByExt = {
  '.svg': () => require(`@talk/img/${filename}.svg`),
  '.png': () => require(`@talk/img/${filename}.png`),
  '.jpg': () => require(`@talk/img/${filename}.jpg`),
  '.jpeg': () => require(`@talk/img/${filename}.jpeg`),
  '.webp': () => require(`@talk/img/${filename}.webp`),
}

if (requiresByExt[ext]) {
  return requiresByExt[ext]()
}
```

ShGKme: > Why are we even limiting it to a set of extensions?

@nickvergessen 

There is a short note in the comment below. A more detailed explanation:

1. This code works in runtime in a Web-Browser. But we have no `require` function in the web browser runtime. 
2. Requiring a module is resolved by the bundler (Webpack) during the bundling. While building the bundle Webpack finds the required module and passes it throw the loader. For files it moves the file to the `dist` directory, renames the file, and returns here the path to the file in the `dist` (or `data:` inline URL for small files)
3. But the path (`@talk/img/${filename}.svg`) is supposed to be evaluated in the runtime... How does it work during the building?
4. Actually, this is not a real `require`, this is a patched by Webpack `require`
5. Webpack sees that there is a literal string as a path. It uses the literal string as the path template (aka glob), imports ALL the files by this path template, and creates a path mapping in the bundle
6. With `require(`@talk/img/${filename}.${ext}`)` template it could literally any file. Webpack will include everything and fail the building if there is anything that cannot be loaded by the current Webpack config.

See also: https://webpack.js.org/guides/dependency-management/#require-with-expression

ShGKme: > Did we come to an agreement? PR with mentioned issue is already on master, so we could proceed here also

I'm not sure if your 👍 was an agreement to not use switch, or to use the mapper object :D

Antreesy: Mapper looks much cleaner and readable to me, so if you have time to change it, would be very cool

Antreesy (approved):
Tested, still works!

ShGKme:
Rebased onto `main` and `rebase --autosquash`, no changes

PR #173, opened 2023-05-05
fix(patchers): fix doubling webroot in generateFilePath

ShGKme:
### ☑️ Resolves

There was a redundant `webroot` in `generateFilePath` patcher. `generateFilePath` already uses `webroot` which breaks some links.

It breaks loading one file: `/core/img/filetypes/file.svg`


Antreesy (approved):
:seal:

Issue #199, opened 2023-05-23
Cannot login with spaces in username


andyR14:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Login with "FirstName LastName" (space/blank between first and last name)
2. Return Message: Login was successful but something went wrong...
3. "One-Word-Usernamens" are working "Name"

### Actual behaviour
![talk](https://github.com/nextcloud/talk-desktop/assets/84279894/5ba7dc02-74a3-4669-8ad9-880cd29c5ce7)

## Desktop client

**Talk Desktop client version:** v0.6.0

**Operating system:** MacOS

**Operating system version:** macOS 13.4

## Server

**Nextcloud version:** 26.0.1

**Talk app version:** 16.0.3

ShGKme:
Thank you for your issue, I'm looking on this.

tobbeuchert:
The same issue appeared to me after installing the .flatpak file on Ubuntu 20.04.6 LTS. Since the issue was closed, what was the solution? Thank you!

PR #202, opened 2023-05-23
fix(auth): fix login for users with space in userid

ShGKme:
### ☑️ Resolves

* Issue #199

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Replace `+` with space ` ` in received `userid` from login flow

P.S. Should it be mentioned in the loginFlow docs?

### DIFF
--- a/src/authentication/login.service.js
+++ b/src/authentication/login.service.js
@@ -42,7 +42,7 @@ function parseLoginRedirectUrl(url) {
 	}
 	return {
 		server: parsed[1],
-		user: decodeURIComponent(parsed[2]),
+		user: decodeURIComponent(parsed[2].replaceAll('+', ' ')),
 		password: parsed[3],
 	}
 }


nickvergessen:
> P.S. Should it be mentioned in the loginFlow docs?

https://www.php.net/manual/en/function.urlencode.php

> Returns a string in which all non-alphanumeric characters except -_. have been replaced with a percent (%) sign followed by two hex digits and spaces encoded as plus (+) signs. It is encoded the same way that the posted data from a WWW form is encoded, that is the same way as in application/x-www-form-urlencoded media type. This differs from the [» RFC 3986](http://www.faqs.org/rfcs/rfc3986) encoding (see [rawurlencode()](https://www.php.net/manual/en/function.rawurlencode.php)) in that for historical reasons, spaces are encoded as plus (+) signs.

That being said, it's also applied to the password:
https://github.com/nextcloud/server/blob/20e00cdf17f45f811135fe5fb61c133ce9021144/core/Controller/ClientFlowLoginController.php#L325

Current default app tokens are not affected, but better be save then sorry.

nickvergessen:
And yes, should be documented in https://docs.nextcloud.com/server/stable/developer_manual/client_apis/LoginFlow/index.html#obtaining-the-login-credentials

nickvergessen:
arg auto-merge

PR #204, opened 2023-05-24
feat(auth): Allow to login with a copy pasted URL from browser

nickvergessen:
### ☑️ Resolves

* Copy paste URL from browser e.g. `https://192.168.178.101/index.php/apps/spreed/` and try to login

Before | After
---|---
grafik | grafik


ShGKme (approved):

ShGKme:
Should we do the same for servers with mod_rewrite (without index.php)?
(Consider the base path to be empty)

nickvergessen:
Didn't do that as it's also not done in the files desktop, or any talk clients.

PR #238, opened 2023-06-22
fix(authentication): Use content size on login web-view to avoid scroll on Windows

ShGKme:
### ☑️ Resolves

* Should not be a regression, but for me it was like a regression of https://github.com/nextcloud/talk-desktop/pull/237

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Use content size for window, so the window size will not include large window shadows on Windows


Antreesy (approved):
Can't check it on Linux, as here is no scroll observed, but according to docs everything seems to be proper :seal: 
https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions

PR #239, opened 2023-06-22
Replace `esm.js` extension to `es.mjs` in nextcloud/auth patch import

Antreesy:
### ☑️ Resolves

* Issue caused by introducing exported output of `nextcloud/auth v2.1.0` https://github.com/nextcloud/nextcloud-auth/pull/516
* which was bumped on spreed master branch https://github.com/nextcloud/spreed/pull/9633
* which lead us to `file is not found`


ShGKme:
Closed in favor of https://github.com/nextcloud/talk-desktop/pull/240

PR #240, opened 2023-06-22
fix(patchers): fix patching modules resolving by require.resolve

ShGKme:
An alternative solution to fix problems mentioned in https://github.com/nextcloud/talk-desktop/pull/239

Previously for patched modules in patching it uses absolute paths to the original modules like
```js
import axios from '../../node_modules/@nextcloud/axios/dist/index.esm.js'

export { getRequestToken } from '@talk/node_modules/@nextcloud/auth/dist/index.esm.js'
```

It breaks if the absolute path is changed inside the module. Mentiond PR fixes the problem for `spreed@stable27+` but breaks for `spreed@stable26`. Because of the absolute path, it may happen with any module at any time.

---

New solution resolves a real path by `require.resolve` and provides it by new aliases, for example
```js
import axios from '@desktop-modules--@nextcloud/axios'

export { getRequestToken } from '@talk-modules--@nextcloud/auth'
```

Also, moved the patchers so their path looks like a path to the corresponding npm module.

---

Thanks @Antreesy to finding and explaining the problem 💙


ShGKme:
Rebased onto main

Issue #198, opened 2023-05-22
Refresh AppData and capabilities without re-login


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce

1. Login in Talk Desktop
2. Update Talk on the server, update Talk settings on the server
3. Get `Nextcloud Talk was updated, you need to reload the page` message
4. Reload the page

Or

1. Login in Talk Desktop and close the client
2. Update Talk on the server, update Talk settings on the server
3. Open the desktop client

### Expected behavior

Talk Desktop has new settings (capabilities) and other server data.

### Actual behavior

Talk Desktop has values that were stored after the login.

- In the first scenario, reload doesn't help.
- In the second scenario, after the application restart, it even doesn't know, that AppData is not valid.
  
  Talk has `x-nextcloud-talk-hash` header. But currently, it is not correct in Talk Desktop. Its initial 
  value is set on the first conversations list load while Talk Desktop stores capabilities in AppData from 
  login.

### Re-request AppData on application start

The app loading is longer.

Not correct according to `Nextcloud Talk was updated, you need to reload the page` message. Should restart the application instead of reloading or changing the message.

### Re-request AppData on UI load

The page loading is longer every time, including new windows and some error handling.

### Store `x-nextcloud-talk-hash` in AppData and re-request on dirty hash

1. Store `x-nextcloud-talk-hash` in AppData, including the dirty flag
2. Request to restart or reload the page on a dirty hash
3. On page/app load check if hash is dirty and re-request only when dirty.

nickvergessen:
> Store x-nextcloud-talk-hash in AppData and re-request on dirty hash

This is what we should do. When a feature is enabled/disabled during the day you don't want to have people complaining that it still shows and doesn't work, just because they would need to manually reload/restart the app.

ShGKme:
> This is what we should do. When a feature is enabled/disabled during the day you don't want to have people complaining that it still shows and doesn't work, just because they would need to manually reload/restart the app.

I'm not sure, I understand you correct (or you understood me :D)

The main purpose of storing the hash is to not block app/window loading with requests each time, but only when it is known to be outdated.

[...]

PR #247, opened 2023-06-26
fix: re-fetch appData on talk hash invalidate

ShGKme:
### ☑️ Resolves

* Issue #198 

### 🖼️ Screenshots

No visual changes

### 🚧 Tasks

`AppData` includes `capabilities` and `userMetadata`.

Before `appData` was collected on the first login, then persist in `localStorage` until logout. It makes it invalid on Talk update or Talk settings update.

This PR:
- Adds talk hash and dirty flag to `appData`
- Adds `appData.service` with functions to re-fetch capabilities and userMetadata
- Adds more utils methods in `AppData`
- Restores `appData` during the **Welcome** screen, without dirty `executeJavascript` in the main process
  - And re-fetch if talk hash is dirty
- Initializes Talk's `talkHashStore` with the hash restored with `appData`
- Subscribe Talk's `setInitialNextcloudTalkHash` mutation to save initial hash
- Subscribe Talk's `markNextcloudTalkHashDirty` mutation to handle invalide hash and relaunch the app

Possible scenarios:

#### Scenario 1 - appData invalidates between launches

1. App restores the appData with the old hash
2. The old hash is set to the Talk's store
3. On Talk app init (conversations loading) it receives new hash
4. Hash is dirty - save it to the appData
5. App is relaunching
6. On relaunch during the welcome screen new capabilities and userMetadata are fetched and persisted

#### Scenario 2 - appData invalidates after the launch

The same

#### Scenario 3 - old appData before this update

The same. Having now `talkHashDirty` flag considers dirty.

## Drawbacks and alternative solutions

1. Show a message instead of the hard relaunch
   - On hash invalidate Talk web shows a toast asking to reload and makes a web page barely working
   - It requires patching Talk web to change the toast behavior
   - But in Talk Web, we cannot call, save drafts or join chats on invalidate anyway
   - Drawbacks:
     - Possibility of saving a draft of the current chats (other drafts are lost anyway)
     - Hard relaunch doesn't look good in UX
   - I prefer the current solution on desktop and think, the game is not worth the candle
2. Always re-fetch new appData on launch
   - Currently, if appData is invalid, it would be known only after launch and fetching conversations
   - It makes the launch fast, but relaunches app on invalidates
   - Alternative - make launch always slower, but without relaunch on invalidate between launches
   - I prefer the current solution because a typical launch doesn't require re-fetch


src/app/appData.service.js
```diff
@@ -0,0 +1,61 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { getCurrentUserData, getCapabilities } from '../shared/ocs.service.js'
+
+/**
+ * Re-fetch capabilities and userMetadata and update appData
+ *
+ * @param {AppData} appData
```
nickvergessen, line 27: Some eslint mimi here and there

src/app/appData.service.js
```diff
@@ -0,0 +1,61 @@
...
+	appData.version.talk = talkCapabilities.version
+	if (persist) {
+		appData.persist()
+	}
+}
+
+/**
+ * If talk hash is dirty, re-fetch capabilities and userMetadata and update appData
+ *
+ * @param {AppData} appData
```
nickvergessen, line 52: Same here

src/authentication/renderer/AuthenticationApp.vue
```diff
@@ -77,6 +77,7 @@ import NcTextField from '@nextcloud/vue/dist/Components/NcTextField.js'
 import { getCapabilities, getCurrentUserData } from '../../shared/ocs.service.js'
```
nickvergessen, line 77: 'getCurrentUserData' is defined but never used

```suggestion
import { getCapabilities } from '../../shared/ocs.service.js'
```

nickvergessen (commented):
Can't read it from the code really, so just commenting here:
In case it is not happening already, we should also refetch the data if the desktop client version changed. So that when people update the client they always get newest data.

ShGKme:
> In case it is not happening already, we should also refetch the data if the desktop client version changed. So that when people update the client they always get newest data.

Do you mean, on any update, even if a talk hash was not changed?

ShGKme:
> yes, if the talk client version changes, refetch the data. Maybe at some point we don't expose a feature to specific client versions or something alike.

Done in the last commit

nickvergessen (approved):
Didn't test, but looks good

ShGKme:
Small update, also save new desktop client version in the appData on re-fetch

PR #254, opened 2023-07-03
Fix incorrect dark theme detection caused by missing css load waiting

ShGKme:
### ☑️ Resolves

Sometimes with a dark theme, some components act like a light theme.

The problem appears when Talk app initializes and starts faster than CSS with the user's theme is loaded.
(Currently, Talk detect dark theme by CSS)

`loadCss` util is supposed to be async and return a promise, but actually, it doesn't.

_This could be not the only problem with dark theme detection thus it also appears out of initial loading._

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 Tasks

- [x] Create and return promise in `loadCss` util

### DIFF
--- a/src/shared/resource.utils.js
+++ b/src/shared/resource.utils.js
@@ -25,21 +25,25 @@ import { appData } from '../app/AppData.js'
  * Load styles from URL via new <link> element
  *
  * @param {string} url - Styles URL
- * @return {HTMLLinkElement} Created styles link element
+ * @return {Promise<HTMLLinkElement>} Created styles link element
  */
-export function loadCss(url) {
-	const link = document.createElement('link')
-	link.rel = 'stylesheet'
-	link.href = url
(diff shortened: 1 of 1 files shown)


Antreesy (commented):
Package seems to work proper, and an error with named import dissapeared, but `npm start` goes into infinite loading for me

ShGKme:
> Package seems to work proper, and an error with named import dissapeared, but `npm start` goes into infinite loading for me

Any errors in the console?

nickvergessen (approved):

nickvergessen:
Right | same here
---|---
grafik | Bildschirmfoto vom 2023-07-04 09-07-57

ShGKme:
Ready to review again after fixing the mentioned issue.

nickvergessen (approved):

PR #272, opened 2023-07-25
fix(tray): use tray only in talk window and re-setup when window changes

ShGKme:
### ☑️ Resolves

Tray sometimes misses the window it is associated with and works inconsistently on different windows:

- Welcome and Login windows have tray icon, but the app doesn't stay in tray when a window is closed
- On logout the main talk window is closed but still exists in the memory as a leaked window process (because the tray keeps the window from being closed)
- After logout opening a login window from the tray crashes the app:
 image

This PR changes the behavior of the Tray. Now the system tray icon is only used in main Talk app window, but is not used in Welcome loading screen and Login window.

### 🚧 Tasks

- [x] Move the `tray` initialization from `app` initialization in `main.js` to `createTalkWindow`
- [x] Destroy the `tray` instance when window is closed
- [x] Keep the tray module to manage `isQuitting` state by itself, not my `main.js`
- [x] Close `talkWindow` with `window.destroy()` on logout to remove it (`close` is prevented by tray)
- [x] Refactor a `setupTray`, use a `browserWindow` as a param instead of click handler, because it always supposed to have the same click handler while setup needs the window to be associated with

### DIFF
--- a/src/app/app.tray.js
+++ b/src/app/app.tray.js
@@ -1,6 +1,8 @@
 /**
  * @copyright Copyright (c) 2023, Julian Petri <julian.petri@iconics.de>
+ *
  * @author Julian Petri <julian.petri@iconics.de>
+ * @author Grigorii Shartsev <me@shgk.me>
  *
  * @license AGPL-3.0-or-later
  *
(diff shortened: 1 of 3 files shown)


ShGKme:
Tray still doesn't work in production build because of icon loading issue. Will be fixed in a separated PR

Antreesy (approved):

PR #273, opened 2023-07-25
Fix tray icon images

ShGKme:
### ☑️ Resolves

Tray icon image loading was broken in packaged build.

### 🖼️ Screenshots

OS | Light | Dark
---|---|---
🪟 | image | image
🐧 | No screenshot, but the same | image
🍎 | image | image 

### 🚧 Tasks

- [x] Add images to the bundle with webpack assets
  - Maybe, it should be `extraResouce`. Then it would require a different solution for development
- [x] Add small 32x32 image for Linux (256x256 is too sharpy)
- [x] Rename Mac icon image for naming consistency
- [x] Add Dark theme icon for Mac

### Alternative solution

We can use mac style icon (without the background) on all platforms.

image

image

@marcoambrosini what do you think?


ShGKme:
P.S. Should be rebased onto main after merging https://github.com/nextcloud/talk-desktop/pull/272

scripts/generate-icons.js
```diff
@@ -31,7 +31,8 @@ const icongen = require('icon-gen')
 async function generateIcons() {
 	const originalPath = path.join(__dirname, '../img/talk-icon-rounded.svg')
 	const originalMacPath = path.join(__dirname, '../img/talk-icon-mac-shadow.svg')
-	const originalMacTrayPath = path.join(__dirname, '../img/talk-icon-plain.svg')
+	const originalMacTrayLightPath = path.join(__dirname, '../img/talk-icon-plain-light.svg')
+	const originalMacTrayDarkPath = path.join(__dirname, '../img/talk-icon-plain-Dark.svg')
```
nickvergessen, line 35: Dark uppercase, light is lower?

nickvergessen (approved):

PR #347, opened 2023-09-12
Fix: catch http errors on development and styles loading

ShGKme:
### ☑️ Resolves

Two related issues:

1. Sometimes start shows just the default background without any visible errors or menu when credentials are not valid anymore or the server is not available on start.

2. Sometimes on development it is not possible to handle if a request was not successful because of an authentication error or other problem.

### 🚧 Tasks

- [x] Add `requesttoken` to the allowed DAV request headers list
- [x] Consider all failed OPTIONS requests as successful in CORS
  - When `OPTIONS` request fails it is always just a general `Network error` for the client. The client cannot determine if it was actually a network error, authentication error or anything else
  - `OPTIONS` request now always returns 200
  - If there was an error, then actual request will fail later with an implicit error
- [x] Handle errors on styles loading
  - It's made a bit hacky. It is not possible to determine style loading error reason. I'm doing an additional general HTTP request to get the error.
  - Then unauthenticated error is handled globally
  - Other errors are shown as an alert, marks app data as invalid to check on the next start and quit the app

#### 🏚️ Before

image

image

#### 🏡 After

- In case of an authentication error: logout as in other cases on invalid authentication
- In case of network/server availability error quit after a message:
  image

### Alternative solutions and follow-ups

1. Cache styles after login to not re-fetch then each init
2. Show a beautiful error instead of alert

### DIFF
--- a/src/talk/renderer/init.js
+++ b/src/talk/renderer/init.js
@@ -22,6 +22,7 @@
 import { register } from '@nextcloud/l10n'
 import { loadServerCss } from '../../shared/resource.utils.js'
 import { appData } from '../../app/AppData.js'
+import { getCapabilities } from '../../shared/ocs.service.js'
 
 /**
  * Initialize Talk application: styles, localization etc.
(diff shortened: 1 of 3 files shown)


src/talk/renderer/init.js
```diff
@@ -35,7 +36,24 @@ export async function init() {
 		loadServerCss('/index.php/apps/theming/theme/light.css'),
 		loadServerCss('/index.php/apps/theming/theme/dark.css'),
 		loadServerCss('/core/css/server.css'),
-	])
+	]).catch(async () => {
+		// It is not possible to determine why styles loading failed in a web-browser
+		// There are 2 the most likely reasons:
+		// 1. Invalid authentication (401 or 500)
+		// 2. Server is available
```
Antreesy, line 43: *unavailable?

Antreesy (approved):
Looks good code-wise

src/talk/renderer/init.js
ShGKme: Thanks, fixed

Issue #278, opened 2023-08-03
Clicking application in dock no longer opens it


juliushaertl:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Use talk desktop on macOS
2. close the window (application keeps running)
3. try to reopen by clicking on the dock icon

### Expected behaviour
The main window appears

### Actual behaviour
Nothing, only the menubar icon brings up the client again, this used to work before https://github.com/nextcloud/talk-desktop/pull/79

## Desktop client

**Talk Desktop client version:** v0.9.0

**Operating system:** MacOS

Note: I left out the remaining info as it seemd to be irrelevant, let me know if I still shall file those.

PR #348, opened 2023-09-12
fix: restore app from system tray on reopen

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/talk-desktop/issues/278

Currently when Talk Desktop executable is run:
- If it wasn't running - it starts
- If it is running - the window is focused
- If it is running but minimized - it is restored and the window is focused
- If it is running but minimized to the system tray and there is no window - nothing happened (it is the bug)

### 🚧 Tasks

- [x] On app focus (including second instance request) - show the window. It also focuses the window and doesn't change anything for a shown window.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -92,7 +92,8 @@ app.whenReady().then(async () => {
 		if (mainWindow.isMinimized()) {
 			mainWindow.restore()
 		}
-		mainWindow.focus()
+		// Show window if it's hidden in the system tray and focus it
+		mainWindow.show()
 	}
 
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Tested on Linux:
- there was the same bug, but with executing a file from Files / Explorer

ShGKme:
> * there was the same bug, but with executing a file from Files / Explorer

Yep, it is not platform-specific, it just didn't restore hidden window

PR #349, opened 2023-09-12
Add release scripts

ShGKme:
### 🖼️ Screenshots

build

_(Steps count are incorrect on the gif, fixed)_

### 🚧 Tasks

- [x] Add script to make a release packages
  - Check that there are no uncommitted changes
  - Clone Talk if needed to `speed`, checkout tag
  - Install all dependencies
  - Package
- [x] Add instructions to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,58 @@ npm run make:windows
 npm run make:all
 ```
 
+## ✈️ Release
+
+1. Create `release/vX.X.X` branch.
+2. Update `CHANGELOG.md`.  
+   1. If a built-in Talk version is to be changed - add a note:
+      ```md
+      ### Build-in Talk update
+
+      Built-in Talk in binaries is updated to $(VERSION) Talk changelog: https://github.com/nextcloud/spreed/blob/master/CHANGELOG.md
+      ``` 
+3. Update `package.json`:  
+   - For minor update:
+     ```sh
+     npm version minor
+     ```
+   - For patch update:
+     ```sh
+     npm version patch
+     ```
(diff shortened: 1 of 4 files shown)


ShGKme:
> only builds the arm64 version, but I guess that's expected on a arm64 processor

Yess. And now you can build an ARM package each release with one command 😀

SystemKeeper (approved):
👍

Antreesy (approved):
Works on Linux
It took some time to clone 'spreed" (have it externally), but i guess, it's cleaner and better for production release purposes, rather then external setup

ShGKme:
A bit not correct for already cloned Talk to `./spreed/`, there is no `fetch/pull`

ShGKme:
> It took some time to clone 'spreed" (have it externally), but i guess, it's cleaner and better for production release purposes, rather then external setup

Cloned repo is not removed, so it will be faster next time.

But at this moment I thought about it if this should be configurable to use talk from your `TALK_PATH` or to use some temp directory...

SystemKeeper:
If you always clone it would make sense to specify `—depth=1` to not clone everything every time

ShGKme:
Updates:
1. Cloning a shallow repo with `--depth=1` only specified tag
2. If is was already cloned - fetch before checkout (also shallow version)
3. Clone to a temp directory so no working tree is touched

SystemKeeper (approved):
Tested and works nicely with the temp directory :-)

ShGKme:
Arr, there is no check for fixup commits on CI in this repo ><

Issue #12, opened 2023-02-06
Add full-featured replacement or use original `@nextcloud/dialogs`


ShGKme:
The current replacement for `@nextcloud/dialogs` is a boring native modal dialog.

PR #407, opened 2023-11-10
fix(patchers): update `OC.dialogs` replacement with `NcDialog`-based adapter

ShGKme:
### ☑️ Resolves

* Fix: #12

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@nextcloud/capabilities": "^1.1.0",
         "@nextcloud/event-bus": "^3.1.0",
         "@nextcloud/l10n": "^2.2.0",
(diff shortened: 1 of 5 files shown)


nickvergessen:
The confirm on the delete should be red like in the web if that is easily doable?🤔

ShGKme:
> The confirm on the delete should be red like in the web if that is easily doable?🤔

I missed that a part of `OC.dialogs` was replaced with `NcDialog` in place in https://github.com/nextcloud/spreed/pull/10812

This screenshot is not actual anymore. Only on replaced `OC.dialogs`.

nickvergessen (approved):

Issue #29, opened 2023-03-06
Run at startup


ShGKme:

ShGKme:
> Did you try adding to systemd?

On Linux I'm not sure, that `systemd` is a good option. I'd go with adding a config to the `~/.config/autostart/`, as it is more general (and simple) way to add a desktop application to startup in many Linux desktop environments with an integration in the desktop environment (as far as I know).

P.S. On Windows and Mac it should be implemented by [Electron's `app.setLoginItemSettings`](https://www.electronjs.org/docs/latest/api/app#appsetloginitemsettingssettings-macos-windows).

PR #428, opened 2023-11-24
feat: add `--background` argv support for opening app minimized to the system tray

### LINKED ISSUE #427: Add option to start in background
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.

---

I have the application inside my autostart entries. This will always start the application with the window being displayed at startup.

Using the `--background` option when starting the application, it should start in the background, id est the application window itself being hidden and only the tray icon being available.

Always closing the application window works to get to the desired state, but a click-less automatic solution would be preferred.

**Additional context**

The regular desktop client already supports the `--background` flag.

ShGKme:
### ☑️ Resolves

* Fix #427
* Will be useful for #29

### 🖼️ Screenshots

Welcome window is not hide. Hiding the welcome window is more problematic because the app is not in the system tray and connection could not be successful.

So for some time it is visible anyway.

bg

### 🚧 Tasks

- [x] Add argv `--background` support: on open if the main Talk window is to be open and there was argv `--background` - do not show the window. It can be opened from the system tray.
- [x] Opening on the login window is not changed

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -34,6 +34,14 @@ const { createWelcomeWindow } = require('./welcome/welcome.window.js')
 const { installVueDevtools } = require('./install-vue-devtools.js')
 
 /**
+ * Parse command line arguments
+ */
+const ARGUMENTS = {
+	// Open Talk window in the background, minimized to the system tray
+	openInBackground: process.argv.includes('--background'),
+}
+
+/**
  * Separate production and development instances, including application and user data
  */
 if (process.env.NODE_ENV === 'development') {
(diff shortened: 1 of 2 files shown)


ShGKme:
> Not sure this is too useful, but as long as it's not default...

It will be useful for run on startup later to not show the window in the center of the screen after reboot 😅

Now it allows to manually set up run on startup.

PR #450, opened 2023-12-07
fix(i18n): Triple to to ellipsis

rakekniven:
Reported at Transifex


ShGKme (dismissed):
Thank you!

The only thing: `…` should be separated with a space.

ShGKme (approved):
Thanks 💙

PR #458, opened 2023-12-13
fix(authentication): system lang separator

ShGKme:
`@nextcloud/l10n` expect `_` separator while OS may return `-`.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
`de-DE` | `de_DE`
image | image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -72,8 +72,8 @@ if (process.env.NODE_ENV === 'production') {
 ipcMain.on('app:quit', () => app.quit())
 ipcMain.handle('app:getOs', () => getOs())
 ipcMain.handle('app:getSystemL10n', () => ({
-	locale: app.getLocale(),
-	language: app.getPreferredSystemLanguages()[0],
+	locale: app.getLocale().replace('-', '_'),
+	language: app.getPreferredSystemLanguages()[0].replace('-', '_'),
 }))
 ipcMain.handle('app:enableWebRequestInterceptor', (event, ...args) => enableWebRequestInterceptor(...args))
 ipcMain.handle('app:disableWebRequestInterceptor', (event, ...args) => disableWebRequestInterceptor(...args))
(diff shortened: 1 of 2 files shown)


ShGKme:
Thanks @dtrieb for catching the issue

nickvergessen (approved):
For inspiration:
https://github.com/nextcloud/server/blob/1bf15da86343b8213a8554fcf3cda78a9000b455/lib/private/L10N/Factory.php#L480-L535

We could also fall back from dialects to base language (e.g. when `de_AT` is configured we could use `de`)

Not sure you can check that, but can also be done in a follow up

ShGKme:
> For inspiration: https://github.com/nextcloud/server/blob/1bf15da86343b8213a8554fcf3cda78a9000b455/lib/private/L10N/Factory.php#L480-L535
> 
> We could also fall back from dialects to base language (e.g. when `de_AT` is configured we could use `de`)
> 
> Not sure you can check that, but can also be done in a follow up

Done

ShGKme:
However, there is a translation issue...

https://github.com/nextcloud/talk-desktop/blob/672cb5a34dd43e71b20e5f56e541ceefda830228/l10n/de.json

nickvergessen:
Right... So either we tell translators that this is not like the normal desktop client and they should translate both, or we remap de_DE to de on translation sync like in the files client

ShGKme:
> Right... So either we tell translators that this is not like the normal desktop client and they should translate both, or we remap de_DE to de on translation sync like in the files client

On https://help.nextcloud.com/t/list-of-resources-and-their-priority-for-translation/78312#mobile-clients-4 it is said

> Follows the language pattern of the web apps, therefore translations should be available for all server supported languages, e.g. de and de_DE

Discussing this with the translator.

ShGKme:
- Rebased onto master to include fixed German translations from translators
- Fixed a small typo: `talk` -> `spreed` in app name in translations

Tested with @Dtrieb

PR #506, opened 2024-02-01
release: v0.21.0

ShGKme:
### Build-in Talk update

- Built-in Talk in binaries is updated to v18.0.3


Antreesy (approved):

nickvergessen (approved):

ShGKme:
`FilePicker` doesn't work since it uses now config from server-rendered initial state. A fix is not very simple, let's release as it is and then release a bug-fix.

PR #537, opened 2024-02-26
fix(patchers): @nextcloud/router@3 compatibility

ShGKme:
### ☑️ Resolves

* Talk Desktop is not possible to build because it is not compatible with `@nextcloud/router@3`

### 🚧 Tasks

- [x] Export `getBaseUrl`
- [x] Change how **globals** are defined. Instead of bundling them via `ProvidePlugin`, actually define globals in `window` same as Server works.
  - [x] Define `window._oc_webroot` + `window._oc_appswebroots` for `@nextcloud/router`


src/patchers/@nextcloud/router.js
```diff
@@ -25,6 +25,10 @@ import { getRootUrl, generateFilePath as _generateFilePath } from '@desktop-modu
 
 export { linkTo, getRootUrl, generateUrl } from '@desktop-modules--@nextcloud/router'
 
+// getBaseUrl added in v3.0.0
+// On Talk Desktop same as getRootUrl, because getRootUrl is absolute already
+export const getBaseUrl = getRootUrl
```
Antreesy, line 30: Now we should be careful about it, I think

Antreesy (approved):
Tested, compiled on Linux correctly

src/patchers/@nextcloud/router.js
ShGKme: Or find a better solution to handle requests and URL...

ShGKme: But I have no ideas

PR #538, opened 2024-02-26
fix(app): allow X-OC-MTIME header for uploading

ShGKme:
### ☑️ Resolves

* It is not possible to upload an image

### 🚧 Tasks

- [x] Allow `X-OC-MTIME` header used in uploading


Antreesy (approved):
Works before and after changes with `package:linux`

ShGKme:
UPD: it only affects dev build

PR #540, opened 2024-02-28
fix: "Share from Nextcloud" support

ShGKme:
### ☑️ Resolves

* FilePicker ("Share from Nextcloud") support. It should have been working after migrating to `@nextcloud/dialogs@5` but after some changes in `@nextcloud/dialogs` it hasn't been working again

### 🚧 Tasks

- [x] Handle initial state for non-existing app (otherwise getting optional Files configs crashed FilePicker)
- [x] Don't delete `requesttocken` header (it allows non-`OCS-APIRequests` requests like getting Files app configs)
  - I have no idea why I deleted this header 😶
  - Actually, it also allows requesting the contact menu for `NcAvatar`
- [x] Allow SEARCH and REPORT requests and X-Requested-With header (otherwise in DEV mode they are blocked by CORS)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,6 @@
 
 - Currently not supported:
   - Screen sharing ([#11](https://github.com/nextcloud/talk-desktop/issues/11))
-  - Share from Nextcloud (including files creation) ([#12](https://github.com/nextcloud/talk-desktop/issues/12))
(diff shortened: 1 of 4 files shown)


Antreesy (approved):
Works, but icons are missing correct width/height:
image

nickvergessen:
Nice side catch! 🎉

DorraJaouad (approved):

PR #542, opened 2024-02-28
fix(talk): support Pinia talkHashStore for Talk 19

ShGKme:
### ☑️ Resolves

Since Nextcloud 29 (Talk 19) 

How to test:
1. Open Talk Desktop
2. Change admin settings
3. In 30 seconds Talk Desktop should restart
4. Should work on `main` / `stable27` / `stable28`

### 🚧 Tasks

- [x] Add Talk Hash Store adapters for Vuex and Pinia
- [x] Check if the Pinia store exists and use Vuex as a fallback
- [x] Work with Talk Hash via adapter

PR #550, opened 2024-03-07
fix(talk): correctly import Vuex store from Talk

ShGKme:
### ☑️ Resolves

* Regression from https://github.com/nextcloud/talk-desktop/pull/542
* Thought, it works fine while importing module with `_esModule: true`...

image


Antreesy (approved):
Lgtm

PR #556, opened 2024-03-11
build: support TS in Talk

ShGKme:
### ☑️ Resolves

* Importing TS with ESM import without extensions in Talk source doesn't work in Talk Desktop, module not found.

```
ERROR in ./spreed/src/components/NewMessage/NewMessage.vue?vue&type=script&lang=js& (./node_modules/esbuild-loader/dist/index.cjs??clonedRuleSet-15!./node_modules/vue-loader/lib/index.js??vue-loader-options!./spreed/src/components/NewMessage/NewMessage.vue?vue&type=script&lang=js&) 25:0-101
  Module not found: Error: Can't resolve '../../services/avatarService'
```

### 🚧 Tasks

- [x] Add TS support in Talk Desktop
- [x] Temporary allow importing TS without extensions
  - This is not a good solution, but spending almost a full working day I have found no other solutions from the Talk Desktop side and I don't understand, why it doesn't work.
  - Alternative - follow server style in Talk and always import with extensions. But it would require a new Talk release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "@nextcloud/eslint-plugin": "^2.1.0",
         "@nextcloud/webpack-vue-config": "^6.0.1",
         "@vercel/webpack-asset-relocator-loader": "^1.7.3",
+        "@vue/tsconfig": "^0.5.1",
         "babel-loader-exclude-node-modules-except": "^1.2.1",
         "babel-plugin-add-module-exports": "^1.0.4",
         "css-loader": "^6.10.0",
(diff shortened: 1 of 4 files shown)


Antreesy:
Fixed that with https://github.com/nextcloud/spreed/commit/4deb3905c1a6683ff4cb9b542996e6c8c9027611

Let's add a rule for speed repo to avoid that in future

Antreesy (approved):

ShGKme:
> Fixed these files with [nextcloud/spreed@4deb390](https://github.com/nextcloud/spreed/commit/4deb3905c1a6683ff4cb9b542996e6c8c9027611)

Yes, but it is not released.

One this will be released, I'll remove this setting.

PR #561, opened 2024-03-15
release: v0.24.0

ShGKme:
### Build-in Talk update

- Built-in Talk in binaries is updated to v19.0.0-beta.2


Antreesy (approved):
There could be some mismatches between current code on server and desktop beta, but should be good)

PR #567, opened 2024-03-20
fix(talk): support original path for root page

ShGKme:
### ☑️ Resolves

* Allows to remove special path alias for welcome page for Talk Desktop: https://github.com/nextcloud/spreed/pull/11877
* This doesn't require support (backport of mentioned PR) for old Talk versions

### 🖼️ Screenshots

No visual changes


Antreesy (approved):

ShGKme:
Note: Alternative solution could be to import `router` and then change the route via `router.replace()` before starting the app. But I felt it more reliable to set initial route before using any other Talk code to make closer to the Web Client.

PR #595, opened 2024-04-03
feat: screen sharing

### LINKED ISSUE #11: Screensharing
Screensharing in Electron requires custom share dialog

ShGKme:
### ☑️ Resolves

* Fix #11

### 🖼️ Screenshots

image

- With live previews (except on Wayland)
- Supports "All screens with audio". All of the other options don't support audio streaming.
- When there is only one screen, the first option is renamed to "Entire screen with audio"
- On Mac - opens settings to allow screensharing if it is not allowed

#### Windows / Mac / Linux (xorg)

https://github.com/nextcloud/talk-desktop/assets/25978914/7f29e4ec-86ca-4597-8e10-cfd5f79527fc

#### Linux (Wayland)

Wayland with PipeWire has a different workflow:
- Requesting all available sources (screens, windows) - triggers native system dialog to choose what to share
- Showing the stream for the selected source also triggers the system dialog
- So live preview for all options is not possible, each preview triggers the dialog
- Instead, only one static image preview is shown for the selected source
- Starting screensharing also triggers the dialog, and the second selected source is actually streamed
- Unfortunately, this is a known issue with double dialog. There is the same issue in Chromium

https://github.com/nextcloud/talk-desktop/assets/25978914/19d3a628-777b-4bf2-ade8-69cf00d5cc01

### 🚧 Tasks

- [x] Add `app:getDesktopCaptureSources` method to get the list of available sources on backend
- [x] Add `DesktopMediaSourceDialog` with the list of available sources to share (screens, windows)
- [x] Add `OCA.Talk.Desktop.getDesktopMediaSource` to invoke the dialog and get the selected source
- [x] On Talk side, use `OCA.Talk.Desktop.getDesktopMediaSource` instead of unavailable `navigator.mediaDevices.getDisplayMedia` (https://github.com/nextcloud/spreed/pull/12003)
- [x] Polishing

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -20,7 +20,7 @@
  */
 
 const path = require('node:path')
-const { app, dialog, BrowserWindow, ipcMain } = require('electron')
+const { app, dialog, BrowserWindow, ipcMain, desktopCapturer, systemPreferences } = require('electron')
 const { setupMenu } = require('./app/app.menu.js')
 const { setupReleaseNotificationScheduler } = require('./app/githubReleaseNotification.service.js')
 const { enableWebRequestInterceptor, disableWebRequestInterceptor } = require('./app/webRequestInterceptor.js')
@@ -28,7 +28,7 @@ const { createAuthenticationWindow } = require('./authentication/authentication.
 const { openLoginWebView } = require('./authentication/login.window.js')
 const { createHelpWindow } = require('./help/help.window.js')
 const { createUpgradeWindow } = require('./upgrade/upgrade.window.js')
-const { getOs, isLinux } = require('./shared/os.utils.js')
+const { getOs, isLinux, isMac } = require('./shared/os.utils.js')
 const { createTalkWindow } = require('./talk/talk.window.js')
 const { createWelcomeWindow } = require('./welcome/welcome.window.js')
 const { installVueDevtools } = require('./install-vue-devtools.js')
@@ -81,6 +81,20 @@ ipcMain.handle('app:setBadgeCount', async (event, count) => app.setBadgeCount(co
 ipcMain.on('app:relaunch', () => {
 	app.relaunch()
 	app.exit(0)
+})
+ipcMain.handle('app:getDesktopCapturerSources', async () => {
+	// macOS 10.15 Catalina or higher requires consent for screen access
+	if (isMac() && systemPreferences.getMediaAccessStatus('screen') !== 'granted') {
+		// TODO: show user-friendly error in this case
+		return []
+	}
+
+	const sources = await desktopCapturer.getSources({ types: ['window', 'screen'], fetchWindowIcons: true })
+	return sources.map((source) => ({
+		id: source.id,
+		name: source.name,
+		icon: source.appIcon?.toDataURL(),
+	}))
 })
 
 app.whenReady().then(async () => {
--- a/src/main.js
+++ b/src/main.js
@@ -28,7 +28,7 @@ const { createAuthenticationWindow } = require('./authentication/authentication.
 const { openLoginWebView } = require('./authentication/login.window.js')
 const { createHelpWindow } = require('./help/help.window.js')
 const { createUpgradeWindow } = require('./upgrade/upgrade.window.js')
-const { getOs, isLinux, isMac } = require('./shared/os.utils.js')
+const { getOs, isLinux, isMac, isWayland } = require('./shared/os.utils.js')
 const { createTalkWindow } = require('./talk/talk.window.js')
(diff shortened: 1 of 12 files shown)


ShGKme:
Rebased onto main to resolve README conflict

nickvergessen:
I receive:
```
Error occurred in handler for 'app:getDesktopCapturerSources': TypeError: systemPreferences.getMediaAccessStatus is not a function
    at /home/nickv/Nextcloud/TalkDesktop/talk-desktop/.webpack/main/index.js:48014:24
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:78183)
    at WebContents.emit (node:events:514:28)
```

On linux

nickvergessen:
Works now, but there is some "stacking issue"

I have to "pick" a screen 3 times. After
1. The "full screen with audio" frame appears
2. The next screen option appears
3. The video in 1. appears

Before 1 | Before 2 | Before 3 | After 3
---|---|---|---
Bildschirmfoto vom 2024-04-05 09-34-07 | Bildschirmfoto vom 2024-04-05 09-34-16 | Bildschirmfoto vom 2024-04-05 09-34-33 | Bildschirmfoto vom 2024-04-05 09-34-41


:speaking_head: At this moment (After 3) I have a Audio Feedback Loop (is it avoidable to play the shared audio?)

After picking either option I have to accept the screenshare again (but that is also the case with normal Chromium)
Bildschirmfoto vom 2024-04-05 09-36-02

After selecting that the screenshare menu is "open" (should be closed)
Bildschirmfoto vom 2024-04-05 09-36-17

ShGKme:
Solution should be different for every OS 🥲

src/main.js
```diff
@@ -82,6 +82,39 @@ ipcMain.on('app:relaunch', () => {
 	app.relaunch()
 	app.exit(0)
 })
+ipcMain.handle('app:getDesktopCapturerSources', async () => {
+	// macOS 10.15 Catalina or higher requires consent for screen access
+	if (isMac && systemPreferences.getMediaAccessStatus('screen') !== 'granted') {
```
nickvergessen, line 87:
```suggestion
	if (isMac() && systemPreferences.getMediaAccessStatus('screen') !== 'granted') {
```

nickvergessen (changes requested):

nickvergessen (approved):

ShGKme:
> > After selecting that the screenshare menu is "open" (should be closed)
> 
> This is still the case, otherwise fine by me

This part is outside of Talk Desktop...

nickvergessen:
I never noticed it in the browser. But seems you are right

Antreesy (approved):
Tested with Linux package, don't have any comments or suggestions. Nice work!

DorraJaouad:
On Windows, I am not sure it is related to the PR or setup 🙈  :
* Icons are not resized
* Call button is not there and after sharing, the sharing button also disappears

image


and I think there should be a watcher over streams so when a the modal is open and you close a window, the list gets updated. For now, it turns into a black stream but remains there.

ShGKme:
> * Icons are not resized

This is `NcDialog` feature... Talk Desktop, and even Talk should have no effect on it. Can you check that you have fresh `npm ci` installed on both `talk-desktop` and `spreed`? And if it is, what styles are applied on the button.

Which version of `spreed` you tested with? 

> * Call button is not there and after sharing, the sharing button also disappears

Also should not be affected by the Talk Desktop...

> and I think there should be a watcher over streams so when a the modal is open and you close a window, the list gets updated. For now, it turns into a black stream but remains there.

Nice catch, checking...

ShGKme:
> I never noticed it in the browser. But seems you are right

I can reproduce in Web:
1. Start a call
2. Start screen sharing
3. Stop the screen sharing
4. Start screen sharing again
5. The menu is open

I have no idea why it happens

ShGKme:
> * Icons are not resized

I cannot reproduce, but after some debugging with @DorraJaouad it seems to be Vue 2.7 + weird v-bind CSS issue. It correctly compiles styles, but it doesn't set CSS variable for "v-bind in css" on the root element of `NcIconSvgWrapper`.

ShGKme:
> and I think there should be a watcher over streams so when a the modal is open and you close a window, the list gets updated. For now, it turns into a black stream but remains there.

Remove preview when the source is not available anymore and request sources each second to check for new sources.

cc @DorraJaouad

ShGKme:
> All good 💯

Even icons? o_O

DorraJaouad:
> Even icons? o_O

No :( , but it's not related to this PR.

PR #603, opened 2024-04-11
fix(patchers): patch router without permanent absolute _oc_webroot

ShGKme:
### ☑️ Resolves

`@nextcloud/router` cannot be used as it is, because it relies on `window.location` which is different on Desktop app.

#### Before

- For functions returning relative link, `window._oc_webroot` was replaced with an absolute base path instead of relative (`https://nextcloud.ltd/cloud` instead of `/cloud`) to return an absolute link
- For functions returning an absolute link, a complete copy-pasting solution was made

It makes it impossible to work with relative links. And it is just not clean

`getBaseUrl` was literally the same as `getRootUrl`.

#### After

`_oc_webroot` is now kept original (relative) and only when needed temporally replaced with an absolute.

Added `OCA.Talk.Desktop.runWithAbsoluteWebroot` helper to run another function having absolute webroot.

`@nextcloud/router` is now patched with this helper and re-defined `baseURL` default value for simplicity, keeping the original implementation.

### 🖼️ Screenshots

Support for relative links in Talk by new autolink in `NcRichText`

🏚️ Before | 🏡 After
---|---
relative-links-before | relative-links-after

Also works with absolute links, but **only if there is no base (webroot)**.

links

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@nextcloud/capabilities": "^1.1.0",
         "@nextcloud/event-bus": "^3.1.0",
         "@nextcloud/l10n": "^2.2.0",
+        "@nextcloud/router": "^3.0.0",
         "@nextcloud/vue": "^8.11.1",
         "core-js": "^3.36.1",
         "electron-squirrel-startup": "^1.0.0",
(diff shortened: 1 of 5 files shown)


Antreesy (approved):
:seal:

Looks valid, but didn't test. Can give it a try, if you describe edge cases that don't work with current app-handling (without PR)

ShGKme:
> Can give it a try, if you describe edge cases that don't work with current app-handling (without PR)

Currently it can be tested with relative links like `[/call/<token>](/call/<token>)`.

But the most important part is that it makes `getBaseUrl` and `getRootUrl` works the same as on Web.
Before on Talk Desktop it was like `getBaseUrl === getRootUrl`, both were absolute.

New approach of patching allows to have much less custom behavior on Talk Desktop, less custom code in the patcher and in the end - to get rid of custom link handlers like https://github.com/nextcloud/talk-desktop/pull/569

Then with https://github.com/nextcloud-libraries/nextcloud-vue/pull/5487 it will cover all the cases. (Currently it misses only links with non-empty webroot).

PR #618, opened 2024-04-17
fix(screensharing): better sort and name available sources

ShGKme:
### ☑️ Resolves

Yesterday during the call, I suddenly selected "Entire screen with audio" option, so other attendants heard themselves. This is not the options I wanted to choose, just mixed-up. And that's me who developed the feature 🙈 

### 🚧 Tasks

- [x] Make a name for the entire screen with audio shorter, and with **Audio** in the beginning
- [x] Change the **icon to just audio** instead of screen with a speaker
- [x] Put this option **after** the screen, not as the first option
- [x] Manually sort sources to always have **screens before windows**

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

Also, sometimes it was randomly in this order before:

image

### DIFF
--- a/src/talk/renderer/components/DesktopMediaSourceDialog.vue
+++ b/src/talk/renderer/components/DesktopMediaSourceDialog.vue
@@ -70,23 +70,26 @@ const requestDesktopCapturerSources = async () => {
 		emit('cancel')
 	}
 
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
LGTM, Tested with two screens

DorraJaouad (approved):

PR #619, opened 2024-04-17
fix: show mime type icon for attachments in messages

ShGKme:
### ☑️ Resolves

Might not be the best solution, but as a workaround.

@Ivansss @mahibi Do you use all available icons for mime types from server, including theming?

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


Ivansss:
Most of them but not all of them.
In Talk iOS we need to update them to the latest version of them https://github.com/nextcloud/talk-ios/issues/1578

src/shared/globals/globals.js
```diff
@@ -24,6 +24,7 @@ import { translate, translatePlural } from '@nextcloud/l10n'
 
 import { appData } from '../../app/AppData.js'
 import { dialogs } from './OC/dialogs.js'
+import { MimeTypeList } from './OC/MimeType.js'
```
Antreesy, line 27: incorrect filename, buiuld is failing

Antreesy (approved):
After fixing the import, works correctly (tested against main and stable27)

DorraJaouad (approved):
Tested 👍🏻 
after filename change in import

mahibi:
> @Ivansss @mahibi Do you use all available icons for mime types from server, including theming?

See https://github.com/nextcloud/talk-android/blob/f5a4274e545f506861dc7baf7f92ec5e73d674f6/app/src/main/java/com/nextcloud/talk/utils/DrawableUtils.kt for android, did not compare..
Some were updated by https://github.com/nextcloud/talk-android/pull/3719

src/shared/globals/globals.js
ShGKme: I though `eslint-pluing-import` should have caught this 🙈

ShGKme:
- Fixed import
- Rebased
- Not using any custom theming from server, mobile clients also don't use them

PR #623, opened 2024-04-18
fix: show main window after login

ShGKme:
### ☑️ Resolves

* Previously, after logging in the main window was minimized to the system tray
* Now it is opened
* No changes for launching app being logged in

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -189,9 +189,7 @@ app.whenReady().then(async () => {
 	})
 
 	const welcomeWindow = createWelcomeWindow()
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Ah, so that's why it happened (noticed recently on Mac machine :eyes:) 

Tested on Linux, seems to be more reliable now. Although there's one nitpick, related to the process, it seems to be broken on server side: 
image

ShGKme:
> Although there's one nitpick, related to the process, it seems to be broken on server side:

Yeah, that's server side

PR #634, opened 2024-04-25
feat: user status

ShGKme:
### ☑️ Resolves

* fix: https://github.com/nextcloud/talk-desktop/issues/26
* fix: user is always offline when using Talk Desktop
* With heartbeat, whether the user is active is defined on `mousemove`/`keydown`/`visibilitychange`. Same timings as on server
* With caching of user status and predefined statuses

### 🖼️ Screenshots

image

image

. | .
---|---
image | image
image | image


src/talk/renderer/UserStatus/components/UserStatusForm.vue
```diff
@@ -0,0 +1,187 @@
...
+const userStatusStore = useUserStatusStore()
+
+/** @type {import('vue').Ref<import('./userStatus.types.ts').UserStatus>} */
+const userStatus = ref({ ...userStatusStore.userStatus })
+
+const backupStatus = ref(userStatusStore.backupStatus ? { ...userStatusStore.backupStatus } : null)
+
+const isDirty = ref(false)
+
+const { predefinedStatuses } = storeToRefs(userStatusStore)
```
Antreesy, line 46: Should be valid as well?

```suggestion
const { userStatus, predefinedStatuses } = storeToRefs(userStatusStore)
```

Antreesy (commented):
I'm confused by the appearance a bit: order and some elements are different from web-version.

Web | Desktop
-- | --
image | image

Although small width is totally enough for the usage. other elements might be reconsidered by us / design team, so we could come up to a common solution for both web / desktop, and extract PR output to a separate library for reuse

src/talk/renderer/UserStatus/userStatus.types.ts
```diff
@@ -0,0 +1,106 @@
...
+	message: null,
+	icon: null,
+	clearAt: null,
+	messageIsPredefined: false,
+	messageId: null,
+}
+
+export type UserStatus = UserStatusCustom | UserStatusPredefined | UserStatusNonCustom;
+
+export type UserStatusPrivate = Omit<UserStatus, 'messageId' | 'messageIsPredefined' | 'statusIsUserDefined'>
```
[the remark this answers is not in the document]
ShGKme: Should be used for the backup status

src/talk/renderer/UserStatus/components/UserStatusForm.vue
ShGKme: No, we need a copy of the store state here. So it can be changed and then saved or not saved.

src/talk/renderer/UserStatus/userStatus.types.ts
```diff
@@ -0,0 +1,106 @@
...
+	type: 'period',
+	time: 'week' | 'day',
+}
+
+export type ClearAtPredefinedConfigForTime = {
+	type: 'end-of',
+	/** The time in seconds until clear */
+	time: number,
+}
+
```
DorraJaouad, line 97: Types are defined in the opposite way (or the handling)

```
if (clearAt.type === 'end-of') {
		switch (clearAt.time) {
		case 'day':
		case 'week':
			return Number(moment(new Date()).endOf(clearAt.time).format('X'))
		}
	}
```

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
```diff
@@ -0,0 +1,90 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { onBeforeUnmount, ref } from 'vue'
+import { useUserStatusStore } from './userStatus.store.js'
+
+/** How often to update the user status */
+const USER_STATUS_UPDATE_INTERVAL = 5 * 60 * 1000 // 5 minutes
+
+/** How log user is considered active */
```
DorraJaouad, line 28:
```suggestion
/** How long user is considered active */
```

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
```diff
@@ -0,0 +1,90 @@
...
+
+	let heartbeatInterval
+	const restartHeartbeat = () => {
+		if (heartbeatInterval) {
+			clearInterval(heartbeatInterval)
+		}
+		heartbeatInterval = setInterval(async () => {
+			try {
+				await userStatusStore.updateUserStatusWithHeartbeat(!isActive.value)
+				// userStatusStore.setUserStatus(updatedStatus)
```
DorraJaouad, line 46:
```suggestion
```

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
```diff
@@ -0,0 +1,90 @@
...
+
+	window.addEventListener('mousemove', markActive, { passive: true, capture: true })
+	window.addEventListener('keydown', markActive, { passive: true, capture: true })
+	document.addEventListener('visibilitychange', handleVisibilityChange, { passive: true })
+
+	onBeforeUnmount(() => {
+		clearInterval(heartbeatInterval)
+		window.removeEventListener('mousemove', markActive)
+		window.removeEventListener('keydown', markActive)
+		document.removeEventListener('visibilitychange', handleVisibilityChange)
```
DorraJaouad, line 86: ?
```suggestion
	onBeforeUnmount(() => {
		clearInterval(heartbeatInterval)
		clearTimeout(isActiveTimeout)
		window.removeEventListener('mousemove', markActive)
		window.removeEventListener('keydown', markActive)
		document.removeEventListener('visibilitychange', handleVisibilityChange)
```

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
```diff
@@ -0,0 +1,90 @@
...
+		if (document.hidden) {
+			isActive.value = false
+		} else {
+			markActive()
+		}
+	}
+
+	window.addEventListener('mousemove', markActive, { passive: true, capture: true })
+	window.addEventListener('keydown', markActive, { passive: true, capture: true })
+	document.addEventListener('visibilitychange', handleVisibilityChange, { passive: true })
```
DorraJaouad, line 80: maybe `touchstart` too ?

src/talk/renderer/UserStatus/userStatus.types.ts
```diff
@@ -0,0 +1,106 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export type UserStatusStatusType = 'online' | 'away' | 'dnd' | 'invisible' | 'offline';
+
+export type UserStatusBase = {
+	/** So-called status type or online status */
+	status: UserStatusStatusType,
+	/** The id of the user */
+	userId: number,
```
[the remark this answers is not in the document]
ShGKme: Done

src/talk/renderer/desktop.app.js
```diff
@@ -20,15 +20,16 @@
  */
 
 import Vue from 'vue'
+import { createPinia, PiniaVuePlugin } from 'pinia'
 import DesktopHeader from './DesktopHeader.vue'
-import { translate, translatePlural } from '@nextcloud/l10n'
-
-Vue.prototype.t = translate
-Vue.prototype.n = translatePlural
```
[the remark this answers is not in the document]
ShGKme: Done

src/talk/renderer/UserStatus/userStatus.service.js
```diff
@@ -0,0 +1,160 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+
+/**
+ * Fetches all predefined statuses from the server
+ *
+ * @return {Promise<import('./userStatus.types.ts').PredefinedUserStatus[]>}
```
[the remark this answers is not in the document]
ShGKme: Yeah, but I didn't want to mix new feature PR with large refactoring/migrating to TS. Only added a types file which is not even used in building.

I'd add TS in a separated PR.

src/talk/renderer/UserStatus/components/UserStatusFormClearAt.vue
```diff
@@ -0,0 +1,108 @@
...
+}
+</script>
+
+<template>
+	<div class="user-status-form-clear-at">
+		<label for="user-status-form-clear-at-input">{{ t('talk_desktop', 'Clear after') }}</label>
+		<NcSelect class="user-status-form-clear-at__select"
+			label-outside
+			input-id="user-status-form-clear-at-input"
+			:clearable="false"
```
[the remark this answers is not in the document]
ShGKme: > also `:searchable="false"`?

Makes sense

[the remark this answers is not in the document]
ShGKme: > it feels weird to not be able to set a custom time range there, but only search through 6 available options

Yeah, reused server style here in the first place. But it would be nice to have a custom time option.

src/talk/renderer/UserStatus/userStatus.types.ts
```diff
@@ -0,0 +1,106 @@
...
+	messageIsPredefined: boolean,
+	/** The id of the message in case it's predefined */
+	messageId: string|null,
+}
+
+export type UserStatusCustomBase = UserStatusBase & {
+	statusIsUserDefined: true,
+	message: string | null,
+	icon: string | null,
+	clearAt: number | null,
```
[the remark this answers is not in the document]
ShGKme: Done

src/talk/renderer/UserStatus/userStatus.types.ts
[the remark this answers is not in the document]
ShGKme: Done

src/talk/renderer/DesktopHeader.vue
```diff
@@ -52,7 +54,11 @@ export default {
 	},
 
 	setup() {
+		const userStatusStore = useUserStatusStore()
+		useUserStatusHeartbeat()
+
 		return {
+			userStatusStore,
```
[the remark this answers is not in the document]
ShGKme: In general, I'd prefer to have it separated:
- Less changes if some values are needed in the setup
- Closer to `SFC Setup` where there is no explicit return but only defined variables

src/talk/renderer/UserStatus/userStatus.types.ts
ShGKme: Fixed

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
ShGKme: Fixed

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
ShGKme: Fixed

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
ShGKme: Fixed

src/talk/renderer/UserStatus/useUserStatusHeartbeat.js
ShGKme: Agree and added some more events.

Also, moved checking for active (idle) to a separate composable.

src/talk/renderer/UserStatus/components/UserStatusForm.vue
```diff
@@ -0,0 +1,187 @@
...
+				:clear-at="userStatus.clearAt"
+				@update:clearAt="patchStatus({ clearAt: $event })" />
+
+			<NcButton class="user-status-form__row"
+				type="secondary"
+				wide
+				:disabled="!statusIsUserDefined"
+				@click="clearUserStatusCustomMessage">
+				{{ t('talk_desktop', 'Clear custom status') }}
+			</NcButton>
```
DorraJaouad, line 118: I think it would be better if it is placed at the bottom because it also clears predefined status.

src/talk/renderer/UserStatus/userStatus.store.js
```diff
@@ -0,0 +1,128 @@
...
+
+	const initPromise = (async () => {
+		await updateUserStatusWithHeartbeat(false, true)
+
+		predefinedStatuses.value = await fetchAllPredefinedStatuses()
+		cachePredefinedStatuses(predefinedStatuses.value)
+	})()
+
+	return {
+		initPromise,
```
DorraJaouad, line 120: It's not used elsewhere

DorraJaouad (approved):
Tested, it works as expected and good code-wise.

ShGKme:
image

nickvergessen:
When the message is empty, clear-after is disabled and has some weird fragment/border in it:
grafik


On default window size the "set user status" button is not visible:
grafik

nickvergessen:
But I'm find with merging and proceeding in a follow up with details

ShGKme:
> When the message is empty, clear-after is disabled

It was done by design, but I though it was clearing only custom status. Now I've found that it also clears online status. Changed to be disabled only when everything is reset.

ShGKme:
Having many concerns from @Antreesy, @DorraJaouad and @nickvergessen about design, reverted it to be closer to the server design:
- Fewer headers, smaller gaps — there is enough space now on the default window's size
- "Clear" is moved down

More changes will be done in a separated PR.

image

ShGKme:
- Rebased onto `main`
- Squashed
- No other changes

Antreesy:
Too late to the party :cry: 

Just checked the PR once more, quite good overall, though have a couple of nitpicks for the follow-up (or WIP re-design):

`Clear custom status` might be tertiary / tertiary-no-bg, as in web, or hidden instead of disabled. Don't say anything about making it wide, as we probably won't fit two buttons in line for all languages:
image

Both lines with a textField seems a bit off grid (see outlined elements):
image

ShGKme:
> or hidden instead of disabled

This would make UI jumping

ShGKme:
> Don't say anything about making it wide, as we probably won't fit two buttons in line for all languages

They are not in one line because they have a different scope. "Clear" is only for a custom status while "set user status" is a whole dialog submit.

ShGKme:
> Both lines with a textField seems a bit off grid (see outlined elements):

Emoji picker button has a different position due to default button styles that have padding if a button has text. If this is an issue, should be fixed on lib side.

"Clear after" is in inline

image

Antreesy:
> "Clear after" is in inline

That should be padded on sides at the same distance, as NcButton padding (as I see it)
image

PR #640, opened 2024-04-29
fix(patchers): get rid of @nextcloud/notify_push patching

ShGKme:
Get rid of `@nextcloud/notify_push` after updating upstream.
- [x] Requires https://github.com/nextcloud-libraries/notify_push-client/pull/7
- [x] Requires `@nextcloud/notify_push@1.2.0` to be published


DorraJaouad (approved):
Removal looks clean 👍🏻

PR #643, opened 2024-05-03
fix(talk): define default outline color for focus visible

ShGKme:
### ☑️ Resolves

* Outline color is a bit unpredictable and may look weird
* Use `var(--color-main-text)` as a default value when it is not defined explicitly 

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image


Antreesy (approved):
Lgtm

PR #657, opened 2024-05-08
refactor(patchers): get rid of @nextcloud/auth patching

ShGKme:
Set the user data on the page, same as Server so `@nextcloud/auth` can get it directly.

See https://github.com/nextcloud-libraries/nextcloud-auth/blob/b0262421614d6136edbefeb3896523e3313d3cae/lib/user.ts#L26


DorraJaouad (approved):

ShGKme:
Rebased onto main to resolve conflict

PR #669, opened 2024-05-22
refactor(patchers): get rid of @nextcloud/axios patching

ShGKme:
### ☑️ Resolves

- Remove unneeded `@nextcloud/axios` patcher
- It was not actually needed at all, because `axios` is stateful module - we can mutate `axios` instance to set it up
- So we only need to do it before starting the app and making other requests


src/shared/setupWebPage.js
```diff
@@ -81,5 +111,6 @@ export async function setupWebPage() {
 	appData.restore()
 	window.OS = await window.TALK_DESKTOP.getOs()
 	applyBodyThemeAttrs()
+	applyAxiosInterceptors()
```
DorraJaouad, line 114: Maybe update the function description above ( for `applyUserData` too)

DorraJaouad (approved):

ShGKme:
- Rebased onto `main` to resolve conflict
- Updated comment

DorraJaouad (approved):

PR #650, opened 2024-05-06
WIP: alternative user status design

ShGKme:
### ☑️ Resolves

* The problem: too many clicks to just go to DnD, for example, to disable notifications. Opening a menu, opening a dialog box, working in it
* The proposal: separate online status and the custom status
* This is also a common practice for other messengers

#### Option 1: sub menu

https://github.com/nextcloud/talk-desktop/assets/25978914/d6e16dd0-4a04-4862-8123-32adbace276e

https://github.com/nextcloud/talk-desktop/assets/25978914/6e2d60f0-d667-4d64-8f3a-619c1b9e2da1

#### Option 2: collapsible menu

https://github.com/nextcloud/talk-desktop/assets/25978914/ba7fccb9-b8d6-4763-9518-bf6d9fbabc21

#### Option 3: collapsible "select" menu

https://github.com/nextcloud/talk-desktop/assets/25978914/1692b126-49b3-4226-a4c3-e644c48bf150

### Option 4: Nextcloud-sub-menu

https://github.com/nextcloud/talk-desktop/assets/25978914/a8400ba2-e67f-4d9a-82a8-96cbd0c98a17

PR #683, opened 2024-06-10
fix(app): improve user menu design

ShGKme:
### ☑️ Resolves

* Separate Menu to User Menu and Main Menu, making both simpler and smaller
* Improve the user menu:
  * Display not only `displayName` but also `userid`
  * Display also server name, link and icon
    * Especially actual for users with multiple instances
  * Improve the user status menu to make it easier to switch to DnD
    * See discussions and drafts in https://github.com/nextcloud/talk-desktop/pull/650
* As a part of the menu rewriting - refactor to SFC Setup

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
. | image
. | image
. | image

### DIFF
--- a/src/talk/renderer/components/UserMenu.vue
+++ b/src/talk/renderer/components/UserMenu.vue
@@ -3,9 +3,70 @@
   - SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
+<script setup>
+import { computed, ref } from 'vue'
+import { storeToRefs } from 'pinia'
+
+import { translate as t } from '@nextcloud/l10n'
+import { generateUrl } from '@nextcloud/router'
+
+import NcAvatar from '@nextcloud/vue/dist/Components/NcAvatar.js'
+import NcPopover from '@nextcloud/vue/dist/Components/NcPopover.js'
+import NcUserStatusIcon from '@nextcloud/vue/dist/Components/NcUserStatusIcon.js'
+
+import MdiCheck from 'vue-material-design-icons/Check.vue'
+import MdiChevronRight from 'vue-material-design-icons/ChevronRight.vue'
+import MdiChevronLeft from 'vue-material-design-icons/ChevronLeft.vue'
+import MdiEmoticonOutline from 'vue-material-design-icons/EmoticonOutline.vue'
+import MdiPencil from 'vue-material-design-icons/Pencil.vue'
+import MdiLogout from 'vue-material-design-icons/Logout.vue'
+
+import ThemeLogo from './ThemeLogo.vue'
+import UiMenu from './UiMenu.vue'
+import UiMenuItem from './UiMenuItem.vue'
+import UiMenuSeparator from './UiMenuSeparator.vue'
+import UserStatusDialog from '../UserStatus/UserStatusDialog.vue'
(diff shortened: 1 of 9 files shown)


src/talk/renderer/components/UserMenu.vue
```diff
@@ -18,54 +79,88 @@
...
+						</UiMenuItem>
+					</template>
+
+					<template v-else>
+						<UiMenuItem tag="a"
+							:href="userProfileLink"
+							target="_blank">
+							<div class="user-menu__name">
+								<strong>{{ user['display-name'] }}</strong>
+								<em>{{ user.id }}</em>
```
Antreesy, line 109: I'm afraid of possible overflow here (not for our instance, but others that use LDAP or another sources with long ids)
image

Antreesy (commented):
Apart from it, looks and works nice! Will check the SFC migration thoroughly later

src/talk/renderer/components/UserMenu.vue
ShGKme: What about swap id <-> View profile?

image

ShGKme: Or remove uid completely...

DorraJaouad: Maybe `text-overflow: ellipsis;` after a certain width ?

src/talk/renderer/components/UserMenu.vue
```diff
@@ -18,54 +79,88 @@
...
+								<template #icon>
+									<span v-if="userStatus.icon" style="font-size: 20px">
+										{{ userStatus.icon }}
+									</span>
+									<MdiEmoticonOutline v-else :size="20" />
+								</template>
+								{{ userStatus.message || t('talk_desktop', 'Set custom status') }}
+								<template #action-icon>
+									<MdiPencil v-if="userStatus.message" :size="20" />
+									<MdiChevronRight v-else :size="20" />
```
DorraJaouad, line 150: Chevron right seems confusing here as It will be a modal open not a submenu.

DorraJaouad (approved):
SFCs are good 🦅

src/talk/renderer/components/UserMenu.vue
ShGKme: Nextcloud Desktop and Talk mobile client doesn't show email/uid anywhere at all.

But I personally would like to keep it.

@nickvergessen what do you think?

src/talk/renderer/components/UserMenu.vue
ShGKme: Removed

src/talk/renderer/components/UserMenu.vue
Antreesy: I'm fine with a place swap only.
Ellipsizing might be tricky given that we don't know possible length of it for different auth methods, and also `display name` + `view profile` would differ from person and locale.

With possible multiple accounts feature (within one app) in future, it will be useful anyway

Antreesy (approved):
Tested, changes look good visually and code-wise 🚀

ShGKme:
- Rebased onto `main`
- Removed `email`/`uid` from first menu item having concerns above

PR #687, opened 2024-06-12
chore(package): update author and license fields to match product license

ShGKme:
### ☑️ Resolves

* `AGPL-3.0` may mean both `AGPL-3.0-or-later` and `AGPL-3.0-only`.
* The product license is `AGPL-3.0-or-later`
* Source files are also in `AGPL-3.0-or-later`
* `package.json` and `package-lock.json` themself are `CC0-1.0`, the `license` field does not define their license.
  https://github.com/nextcloud/talk-desktop/blob/ca8b1138d684bcb43f5a9d749f05b8f700e07bab/.reuse/dep5#L20-L23
* Also changed `author` format and email to match `AUTHORS.md`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
     "": {
       "name": "talk-desktop",
       "version": "0.29.0",
-      "license": "AGPL-3.0",
+      "license": "AGPL-3.0-or-later",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
Looks valid. Noticed that in 'About' section of app, so thought it should reflect the source code license

PR #716, opened 2024-07-16
refactor: migrate to Composition API with SFC Setup and new defaults

ShGKme:
Follow 2022+ Vue recommendations and defaults:

- Migrate to Composition API
  - Using SFC Setup as a new default recommendation and a place with better TS support with new features in Vue 3
- Change block order to `script, template`
  - Places template close to styles

Add corresponding ESLint rules

### DIFF
--- a/src/help/renderer/HelpApp.vue
+++ b/src/help/renderer/HelpApp.vue
@@ -6,19 +6,19 @@
 <template>
 	<div class="about">
 		<h2>{{ t('talk_desktop', 'About') }}</h2>
-		<p>{{ $options.packageInfo.productName }} - {{ $options.packageInfo.description }}</p>
+		<p>{{ packageInfo.productName }} - {{ packageInfo.description }}</p>
 		<ul class="about__list">
 			<li>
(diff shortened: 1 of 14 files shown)


src/help/renderer/HelpApp.vue
```diff
@@ -38,73 +94,6 @@
...
-export default {
-	name: 'HelpApp',
-
-	components: {
-		MdiWindowClose,
-		NcButton,
-		NcTextArea,
-	},
-
-	inheritAttrs: false,
```
Antreesy, line 58: Not needed here anymore?

For myself: https://github.com/vuejs/rfcs/blob/master/active-rfcs/0040-script-setup.md#declaring-additional-options

Antreesy (approved):
Apart from it, migrated 1:1

src/help/renderer/HelpApp.vue
ShGKme: It was here by mistake. Added a "long" time ago, then Help was implemented as a `NcModal` wrapper.

In SFC Setup it's easy to add it via
```js
defineOptions({
  inheritAttrs: false,
})
```

But in Vue 2 it is not available and requires an additional block:
```html
<script>
export default {
  inheritAttrs: false, 
}
</script>
```

PR #737, opened 2024-08-02
fix(UserMenu): minor visual issues

ShGKme:
### ☑️ Resolves

* There was a focusable element inside another focusable element
  * Some server global styles from button also have high specificity
  * Made avatar a button instead of wrapping into HTML button
  * Opens manually now, because floating vue doesn't support `role=button` as buttons
* Submenu is kept open after reopening the user menu
  * Closed
* Menu has unlimit width
  * Limited
* Server icon is not a square
  * Adjusted size
* Title bar has server specificfocus-visible underline
  * Removed

### 🖼️ Screenshots

🏚️ Before (2 focus) | 🏡 After (1 focus)
---|---
image | image
image | .

🏚️ Before | 🏡 After
---|---
image | image
before | after
image | image
image | image

### DIFF
--- a/src/talk/renderer/assets/overrides.css
+++ b/src/talk/renderer/assets/overrides.css
@@ -18,3 +18,13 @@ body {
 	margin-top: var(--header-height) !important;
 	border-radius: 0 !important;
 }
+
+/* Remove Nextcloud header specific focus-visible style */
+#header {
+	.button-vue,
+	[role=button] {
+		&:focus-visible::after {
+			content: none !important;
+		}
+	}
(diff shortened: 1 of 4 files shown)


ShGKme:
Adding one more small change

Antreesy (approved):
Tested

There is one `::after` element on MainMenu could be removed
image

ShGKme:
> There is one `::after` element on MainMenu could be removed

Fixed

ShGKme:
Rebased onto `main`

PR #741, opened 2024-08-03
feat: use local server styles instead of requested

ShGKme:
### ☑️ Resolves

Currently Talk Desktop has copy of `master` server styles used on the Login and then request remote server styles from the connected Nextcloud server.

Problems:
- Talk frontend in the lastest version might not be compatible with old Nextcloud server
- Loading Talk window is slower, waiting for the styles request

Proposal:
- Store all styles locally
- Add an util to get the styles
- To make it compatible with different server versions - store them separately

New indirectly added features:
- All windows no support the dark theme

Drawbacks:
- Custom themes, including primary color and background, are not supported anymore :(

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
before-load | after-load

### 🚧 Tasks

- [x] Add script to fetch server styles for a specific Server version
  - Copies core styles and files
  - Fetches PHP generated server styles from themes app
  - Adjusts .reuse/dep5 for fetched styles
- [x] During the build, get the styles depending on used Talk version
- [x] Import local styles in app instead of making the request
- [x] Adjust styles a bit


Antreesy (approved):
Tested with linux package. Feels a lot, but make sense to store the styles locally

ShGKme:
Note: currently it doesn't support `stable29` and below, because it relies on `.reuse/dep5` introduced in Nextcloud 30. but we aren't planning to built Talk Desktop with 29 anymore

ShGKme:
It's only the first stage on updating styles workflow.
Next steps:
- automatically request switch and request required styles on build
- improve fetching script to make it faster and reusable

Merging for now, seems to work. Huge change is only around generated files.

PR #742, opened 2024-08-03
fix(patchers): ignore unset baseURL in @nextcloud/router

ShGKme:
### ☑️ Resolves

* Talk Desktop defines the default value for `baseUrl` in `@nextcloud/router` requests, it is it not explicitly set
* But if it's set as `undefined` or other `falsy` value, Nextcloud Talk keeps it as a set value
* It results in requesting without `baseURL`
* Solution: only set `baseURL` if it is not actually set with `truthy` value

Example of the problematic place:
https://github.com/nextcloud/spreed/blob/29af6dac6325af9bba16b4706bb09066b601165e/src/utils/signaling.js#L980-L981


Antreesy (approved):
And signaling is back working!

PR #746, opened 2024-08-09
fix: make Talk window more compact to fit into full HD

ShGKme:
### ☑️ Resolves

* Previous 1680x1050 was too huge for FHD screen (1050 + title bar > 1080 - task bar)
  * With a new compact design we can make it smaller and yet fit Media Settings
  * Ref: https://github.com/nextcloud/talk-desktop/pull/682
* ~~Make the window initially maximized~~
  * ~~It seems like most of the users use it maximized anyway~~
  * Dropped

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/src/talk/talk.window.js
+++ b/src/talk/talk.window.js
@@ -44,8 +44,8 @@ function createTalkWindow() {
 
 	const window = new BrowserWindow({
 		...talkWindowOptions,
-		width: Math.min(1680, screenWidth),
-		height: Math.min(1050, screenHeight),
+		width: Math.min(1400, screenWidth),
+		height: Math.min(900, screenHeight),
 		show: false,
 	})
 


src/talk/talk.window.js
```diff
@@ -44,11 +44,14 @@ function createTalkWindow() {
...
 		...talkWindowOptions,
-		width: Math.min(1680, screenWidth),
-		height: Math.min(1050, screenHeight),
+		width: Math.min(1400, screenWidth),
+		height: Math.min(900, screenHeight),
 		show: false,
 	})
 
+	// Maximize window initially
+	window.maximize()
```
nickvergessen, line 53: Maybe only when screen size is not 4k?

nickvergessen (commented):
*me cries in 4k*

src/talk/talk.window.js
ShGKme: I can drop the commit (´。＿。｀)

Not sure, I want to make it depend on some the screen size. Even on 4k I personally always open it maximized

nickvergessen: yeah lets not do that then, until we store/restore the window size/position in the future :)

ShGKme: Dropped

nickvergessen (approved):

PR #761, opened 2024-08-20
fix(talk): set --header-height

ShGKme:
### ☑️ Resolves

* The header was still visually 50px even its content and system title bar is 50px. It should be 46px.
  * On screenshots see 4px gap under the minimize button
* By mistake I set height in `initLocalStyles` which is not used after another PR made that time

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


Antreesy (approved):
LGTM, can't test on Linux

Do we need to export a function, if it's used in the same file? I don't see it can be needed somewhere else

ShGKme:
> Do we need to export a function, if it's used in the same file? I don't see it can be needed somewhere else

Removed export

PR #766, opened 2024-08-23
fix: handle system theme change

ShGKme:
### ☑️ Resolves

Preparation to complete theme change support.

* Before, there was no support for the `default` Nextcloud theme (= system preferences)
  * There where only `light` and `dark` theme
  * App checked system preferences only on init
* Now:
  * Also load so called `plain` theme styles with `prefers-color-scheme` like the server does
  * Initially set `default` theme instead of specific
  * This allows handling system theme change in a native way like server does

### 🖼️ Screenshots

https://github.com/user-attachments/assets/5963216b-f482-4c18-bfc1-39096afe7318


ShGKme:
Note: avatar change is a new feature of `spreed` PR

Antreesy (approved):
Tested, really nice!

Don't you want to add a toggle to control it locally in client?

ShGKme:
> Don't you want to add a toggle to control it locally in client?

I do, it's several separated PRs.

Just found this issue in progress. Debugging system theme switch didn't work in Talk Desktop

PR #799, opened 2024-09-14
fix(state): get typing privacy from capabilities

Antreesy:
### ☑️ Resolves

* Fix disabled typing indicator in conversations

### 🚧 Tasks

- [ ] Follow-up / Fixup: 
	- [ ] Desktop client doesn't refetch capabilities, if they are changed (but hash isn't dirty)


ShGKme (approved):
🙈

Issue #27, opened 2023-03-06
Desktop settings


ShGKme:
- [ ] Add backend API for storing, reading and updating application-level settings
- [ ] Add IPC API to use settings in the frontend
- [x] Add Talk Frontend API to add new Talk settings sections
- [ ] Add components and modules to work with settings on the Talk Desktop frontend

Follow-up:

- [ ] Implement settings:
  - [ ] General:
    - [ ] Run on startup
  - [ ] Appearance:
    - [ ] Theme
    - [ ] Use custom title
    - [ ] Spellcheck languages
  - [ ] Notifications:
    - [ ] Enable notifications
    - [ ] Play custom Talk notification sounds or system notification sound
  - [ ] Privacy:
    - [ ] Show/hide message content in a notification
    - [ ] Show/hide message author in a notification

PR #835, opened 2024-10-21
feat: add application config aka settings

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/27

### 🖼️ Screenshots

image

### 🚧 Tasks

Separated by commits with 💙

- [x] Add `AppConfig` - application level config
  - Stored in a file
  - Not bind to a specific user session
- [x] Add frontend for Settings, which includes:
 - Frontend API to init the config
 - Store and composables to work with AppConfig
 - Add a Talk Settings Section "Application"
- [x] Add some simple settings

More settings are ready and will come in a separate PR

### DIFF
--- a/src/app/AppConfig.ts
+++ b/src/app/AppConfig.ts
@@ -1,0 +1,144 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { join } from 'node:path'
+import { readFile, writeFile } from 'node:fs/promises'
+import { app } from 'electron'
+import { isLinux } from '../shared/os.utils.js'
+
+const APP_CONFIG_FILE_NAME = 'config.json'
+
+// Windows: C:\Users\<username>\AppData\Roaming\Nextcloud Talk\config.json
+// Linux: ~/.config/Nextcloud Talk/config.json (or $XDG_CONFIG_HOME)
+// macOS: ~/Library/Application Support/Nextcloud Talk/config.json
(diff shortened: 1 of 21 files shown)


src/app/AppConfig.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Get an application config value
+ * @param key - The config key to get
+ * @return - If key is provided, the value of the key. Otherwise, the full config
+ */
+export function getAppConfig<T extends keyof AppConfig>(key?: T): AppConfig | AppConfig[T] {
+	if (!initialized) {
+		throw new Error('The application config is not initialized yet')
+	}
+
+	const config = Object.assign({}, defaultAppConfig, appConfig)
```
Antreesy, line 122: Not sure if there's more performant-wise way to do that, rather than creating a new object on each call
Though it's called just once per window, so shuld be fine?

Antreesy (approved):
Tested, code-wise good.
Maybe Select model part is a bit overcomlicated, and we should upstream it instead?

src/app/AppConfig.ts
ShGKme: > once per window

No, once per app start.

And these are small objects, creating a new object here is not measurable performance-wise. comparable to objects we create in many HTTP calls, computeds and ect.

This code is equivalent to
```js
const config = {
  ...defaultAppConfig,
  ...appConfig,
}
```

Antreesy: Then it's fine, I guess. We can come back to it, when config will grow to enormous size

ShGKme: I don't think the config could be so large that creating a new object here would ever be comparable to inter-process communication. Only if it has thousands of properties.

DorraJaouad (approved):
Code-wise 👍🏻 
Yet, I didn't test because of some setup issues :(.

Issue #824, opened 2024-10-04
Feature Request: Direct File Download in Nextcloud Talk on Desktop


Morteza-j8:
Is your feature request related to a problem? Please describe. I am frustrated that when I send files through Nextcloud Talk, it opens a browser for downloading instead of allowing direct downloads or viewing within the application. This makes the process cumbersome and less efficient for users on both Linux and Windows.

Describe the solution you'd like I would like to see a feature that enables direct downloading and viewing of files shared within the Nextcloud Talk application. This would enhance user experience by allowing seamless interaction without the need to switch to a browser for file downloads.

Describe alternatives you've considered Currently, the only alternative is to use the browser to download files, which disrupts the flow of communication in Nextcloud Talk. Another option is to use other messaging platforms that allow direct file downloads, but I prefer to keep using Nextcloud for its other features.

Additional context Implementing this feature would make Nextcloud Talk a more competitive option for communication, as many users are looking for efficient file-sharing capabilities without unnecessary steps.

ShGKme:
Makes sense

PR #840, opened 2024-10-23
feat: handle download links 

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/824
* Requires: https://github.com/nextcloud/spreed/pull/13618

Unfortunately, there is no better way to handle download links. Electron doesn't handle it, doesn't trigger `will-download` nor allows detecting that the link was a download link in `will-navigate`

### 🖼️ Screenshots

image

### DIFF
--- a/src/app/applyDownloadNotification.js
+++ b/src/app/applyDownloadNotification.js
@@ -26,14 +26,14 @@ export function applyDownloadNotification(browserWindow) {
 				notification.on('click', () => {
 					shell.showItemInFolder(pathToFile)
 				})
-			} else {
+			} else if (state === 'interrupted') {
 				notification = new Notification({
 					title: 'Download Failed',
 					body: `Something went wrong with the download of '${base}'.`,
 				})
 			}
 
-			notification.show()
+			notification?.show()
 		})
 	})
 }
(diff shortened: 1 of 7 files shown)


Antreesy (approved):
Tested

If `download` attr doesn't specify file extension, it's not considered for file name suggestion:
image

Something to fix in Talk only?

ShGKme:
> If `download` attr doesn't specify file extension, it's not considered for file name suggestion:

Is there an example where we have it?

Poll export is `.json`

image

And for me it works fine in both cases:
- download value is without extension
- file url is without extensions

ShGKme:
Checked with @Antreesy.

The problem is that `download` attribute value is ignored, and neither inline `data:*` URL, nor `URL.createObjectURL` allows explicitly setting a filename. As the result, suggested filename is either `download.json` or UUID from `objectURL` instead of the `download` attribute.

Antreesy (commented):
Remarkable workaround 🙈 
For ~~me~~ the record, file extension should be always provided
image

Antreesy (approved):

ShGKme:
> Tested. For ~me~ the record, file extension should be always provided

filename is defined in the same as in web-borser - via the `download` attribute. The extension must be provided there.

Issue #845, opened 2024-10-25
Nextcloud Talk window cannot be reopened after closing on macOS


petarmarj:
## Steps to reproduce
1. Open Nextcloud Talk on Mac 
2. Close the window (eg. with Command + W)
3. Be unable to reopen the window 

### Expected behaviour
The window should reopen.

### Actual behaviour
The window stays closed. But I see the Talk application open and I see the Mac menu bar with Talk application options. There is no way to reopen the window without quitting and reopening the application.

## Desktop client

**Talk Desktop client version:** v0.39.0

**Operating system:** MacOS

**Operating system version:** tested on many versions (today on 10.15.7 and 12.7.6).

**Microphone available:** not relevant

**Camera available:** not relevant

## Server

**Nextcloud version:** 29.0.4.2

**Talk app version:** 19.0.10

**Custom Signaling server configured:** no

**Custom TURN server configured:** no

**Custom STUN server configured:** no

PR #852, opened 2024-10-26
fix(app): show window on activate on macOS

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/845

Unlike other OS, when all windows are closed on macOS, the app is still kept running (app's menu bar is active, icon in the Dock looks like open's app icon).

New click on the Dock icon should "activate" the app and create the window again.

But the window is created hidden by default (stays in the system tray).

The fix - explicitly show the window on activate.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -283,6 +283,7 @@ app.whenReady().then(async () => {
 	app.on('activate', () => {
 		if (BrowserWindow.getAllWindows().length === 0) {
 			mainWindow = createMainWindow()
+			mainWindow.once('ready-to-show', () => mainWindow.show())
 		}
 	})
 })


src/main.js
```diff
@@ -283,6 +283,7 @@ app.whenReady().then(async () => {
 	app.on('activate', () => {
 		if (BrowserWindow.getAllWindows().length === 0) {
 			mainWindow = createMainWindow()
+			mainWindow.show()
```
Antreesy, line 286: Would it make sense to wrap mainWindow into `once` listener, like in other places? Or it won't work because of `app.on` / MacOS specific handling?

```suggestion
			mainWindow.once('ready-to-show', () => mainWindow.show()))
```

Ivansss:
> does this match the MacOS user expectation?

Yes, clicking on the Dock icon and create and show the window again would match user expectations.

Antreesy (approved):

Issue #793, opened 2024-09-12
Add configuration option/screen for zoom


kesselb:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**

Follow-up for / related to: https://github.com/nextcloud/talk-desktop/issues/74

You can change the zoom level with Ctrl +/- or by using the menu.

A nice enhancement would be to show an "accessibility" option in the settings to configure the zoom like Slack does.

**Describe the solution you'd like**

I would like a configuration option to adjust the zoom level for the application.

**Describe alternatives you've considered**

**Additional context**

![image](https://github.com/user-attachments/assets/a6d0a847-a115-47b8-9cb0-602203cd7c09)

PR #853, opened 2024-10-28
feat: add zoom config on UI and fix zoom-related issues

ShGKme:
## ☑️ Resolves

* Fixes: https://github.com/nextcloud/talk-desktop/issues/793
* Adds `zoomFactor` config to application config
* To handle zoom change from different sources, adds listening to config changes feature
* Adds zoom config to the UI
* Considers zoom in **title bar height** and **initial/min window size**

#### Solution complexity

Zoom can be changes with user settings
- Via new user settings
- Via <kbd>CTRL + 🖱️</kbd>/**Touchpad gesture** - implemented by handling the `zoom-changed` event
- Via <kbd>CTRL + +/-</kbd> shortcut - included to the standard corresponding `MenuItem`s and cannot be handled

Because the last option is not possible to handle in Electron at all, and there is no generic zoom changed event, the solution is a bit complex.

#### Zoom step

On each step zoom can be changed:
1. Adding a percent, e.g. `+20%`
   - Scale on level N = 1 + 0.2 * N
   - `20% - 40% - 60% ... 100% - 120% ... 300% - 320%`
   - Simple and understandable for a user
   - Non liner. Change `20% -> 40%` is dramatic `x2` while `300% - 320%` slightly visible
2. Applying a percent, e.g. `x20%`: 
   - Scale on level N = 1.2 ** N
   - `100% - 120% - 144% - 172.8%`
   - Uniform increase
   - Default Electron behavior
   - Hard to understand by user and not beautiful numbers
3. Having a specific pre-defined list of values
   - More complex implementation
   - Beautiful and yet uniform

Option 2 was chosen.

### Settings

settings

### Title bar height change (pay attention on the window buttons)

🏚️ Before | 🏡 After
---|---
before | after

### Windows on 200%

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image
image | image


src/app/AppConfig.ts
```diff
@@ -59,19 +64,26 @@ export type AppConfig = {
 	// Nothing yet...
 }
 
+export type AppConfigKey = keyof AppConfig
```
Antreesy, line 67: doesn't seem to make much difference for me, also TSC doesn't complain without it

src/talk/renderer/Settings/DesktopSettingsSection.vue
```diff
@@ -66,6 +86,40 @@ function relaunch() {
...
+						<template #icon>
+							<IconMinus :size="20" />
+						</template>
+					</NcButton>
+					<NcTextField :id="inputId"
+						class="zoom-input"
+						:aria-describedby="descriptionId"
+						inputmode="number"
+						:value="zoomFactorPercentage"
+						@change="zoomFactorPercentage = $event.target.value" />
```
Antreesy, line 109: Not a big deal, but should we handle border values here (NcTextField, NcButton)?

Antreesy (approved):
Tested on Win

src/talk/renderer/Settings/DesktopSettingsSection.vue
ShGKme: It is handled by `zoomFactorPercentage`

src/app/AppConfig.ts
ShGKme: Added usage of it in other places

src/talk/renderer/Settings/DesktopSettingsSection.vue
Antreesy: I meant that it's still possible to set out of range values in UI, they just not kept in config and not reflected in the app window

src/talk/renderer/Settings/components/SettingsSelect.vue
```diff
@@ -33,44 +35,24 @@ export default {
...
+			<slot name="icon" :size="size" />
+		</template>
+
+		<template #default="{ inputId }">
+			<NcSelect v-model="model"
+				class="settings-select"
+				:options="options"
+				:clearable="false"
+				:searchable="false"
+				:input-id="inputId" />
```
DorraJaouad, line 49: image

ShGKme: Fixed

src/app/zoom.service.ts
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// Limit zoom level to custom title bar limits
+	if (zoom < 0.55 || zoom > 5) {
+		return
+	}
+
+	browserWindow.webContents.setZoomFactor(zoom)
+	// Resize the title bar to match the new zoom level
+	browserWindow.setTitleBarOverlay({
```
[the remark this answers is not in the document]
ShGKme: Fixed

src/talk/renderer/Settings/DesktopSettingsSection.vue
ShGKme: Fixed

Antreesy: Yup, works better with manual input. Maybe same getter/setter could be done for `zoomFactor`?
2024-10-29_16h43_11

Antreesy:
After hard restart window size will be 100%, regardless of config value
B | A
-- | --
image | image

Also label next to NcSelect is overlayed

src/talk/renderer/Settings/components/SettingsSelect.vue
Antreesy: image

src/talk/renderer/Settings/DesktopSettingsSection.vue
ShGKme: It has different limits. Will put it to constants

DorraJaouad (approved):
Tested ✅ 🦅

ShGKme:
> Also label next to NcSelect is overlayed

This part I cannot reproduce

Antreesy:
> This part I cannot reproduce

Styles aren't scoped to NcDateTimePicker: https://github.com/nextcloud-libraries/nextcloud-vue/pull/6095
image
cc @GretaD

Antreesy (approved):
So bug is unrelated to this PR, new changes look good

ShGKme:
Squashed fixups, rebased onto main

PR #854, opened 2024-10-28
fix(notifications): remove notification (talk) icon from notification image

ShGKme:
### ☑️ Resolves

In web-browser it makes sense to show Talk logo as notification image.

But on Desktop all Talk Desktop notifications are from the Talk, and there is an application icon already.

Icon show as the notification **image** should be removed.

Also, it is always dark.

image


Antreesy (approved):
Tested with windows Squirell build

PR #859, opened 2024-10-30
fix(app): handle opening new installation instance

ShGKme:
### ☑️ Resolves

- Allows launching a new instance of Talk instead of a running one
- Prevents a situation when there is an old Talk keep running in the background instead of a new one after installation 

### How to test

1. Build executable with `npm run build`
2. Copy it to a new location. now you have 2 "installations"
3. Run the first one, open some chat/settings
4. Minimize or close the window
5. Try to run the first installation again
   - It should open/focus the window
6. Open the second installation
   - **Before:** the first window is open instead of the new installation
   - **After:** the first installation is closed and the new is launched
7. Run the second installation again
   - Just to make sure it still works as expected for a single instance

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -4,6 +4,7 @@
  */
 
 const path = require('node:path')
+const { spawn } = require('node:child_process')
 const { app, dialog, BrowserWindow, ipcMain, desktopCapturer, systemPreferences, shell } = require('electron')
 const { setupMenu } = require('./app/app.menu.js')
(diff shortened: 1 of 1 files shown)


Antreesy (approved):

Antreesy:
> 1. Build executable with npm run build
> 2. Copy it to a new location. now you have 2 "installations"

That's why it doesn't work with old builds 👀

ShGKme:
Rebased onto `main`

ShGKme:
> That's why it doesn't work with old builds 👀

Not sure what you mean

Issue #292, opened 2023-08-11
Notifications: add popup window for a call notification with join/decline buttons


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**

In contrast to a text message, a call notification often needs an immediate reaction to join a call.

**Describe the solution you'd like**

Add a pop-up window with a call notification with two buttons:
- "Join" - opens the talk window, opens the chat, joins the call
- "Decline" - closes the popup

**Describe alternatives you've considered**

Use native notifications. Make a call native notification always shown and add buttons to the notification.
It is definitely possible on Windows, but I'm not sure it is possible on MacOS and all the Linux desktop environments.

**Additional context**

We have it on both Android and iOS mobile clients. See also:

DorraJaouad:
Is it shown when Talk desktop is minimized and/or when it is expanded ?

nickvergessen:
Both

PR #868, opened 2024-11-04
feat: Call notification popup

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/292

### 🖼️ Screenshots

call

### 🚧 Tasks

- [x] Add callbox window
- [ ] Reuse AvatarWrapper from Talk
- [ ] Open window on call notification
- [ ] Join call with direct call


ShGKme:
Spacing between the avatar and caller name is increased (not updated on screenshots yet)

image

jancborchardt (changes requested):
Very nice! Just some small feedback:
- [ ] Considering the vertical whitespace between caller and buttons it possibly looks better to just make the modal square (I do appreciate your use of the golden spiral though ;)
- [ ] In Nextcloud and also on other calling systems like on iOS, Discord, etc the "Join call" button is usually on the right while "Decline" is on the left

ShGKme:
> * In Nextcloud and also on other calling systems like on iOS, Discord, etc the "Join call" button is usually on the right while "Decline" is on the left

On all the systems or only on Apple?

src/callbox/callbox.window.ts
```diff
@@ -0,0 +1,72 @@
...
+		webPreferences: {
+			preload: CALLBOX_WINDOW_PRELOAD_WEBPACK_ENTRY,
+			enablePreferredSizeMode: true,
+		},
+	})
+
+	window.setPosition(Math.round((screenWidth - width) / 2), Math.round(height / 2))
+
+	params = {
+		token: 'ot4t92uh',
```
nickvergessen, line 60: todo :upside_down_face:

src/callbox/renderer/callbox.html
```diff
@@ -0,0 +1,15 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+  -->
+
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8" />
+	<title>Incoming call</title>
```
nickvergessen, line 10: Translate?

src/main.js
```diff
@@ -261,6 +263,13 @@ app.whenReady().then(async () => {
 			const isTalkWindow = createMainWindow === createTalkWindow
 			if (!isTalkWindow || !ARGUMENTS.openInBackground) {
 				mainWindow.show()
+				createCallboxWindow(mainWindow, {
+					token: 'ot4t92uh',
```
nickvergessen, line 267: todo

src/talk/renderer/notifications/notifications.store.js
```diff
@@ -157,27 +164,49 @@ export function createNotificationStore() {
...
+				token: notification.objectId,
+				name: notification.subjectRichParameters.name,
+				type: notification.subjectRichParameters['call-type'],
+				avatar: notification.subjectRichParameters['icon-url'],
 			}
-			window.TALK_DESKTOP.focusTalk()
-			// Talk will open the call from notification if necessary
-			emit('notifications:action:execute', event)
-		}, false)
+			// const params = {
```
nickvergessen, line 180: todo

src/callbox/callbox.window.ts
Antreesy: Should still be there?

src/talk/renderer/notifications/notifications.store.js
```diff
@@ -355,3 +378,14 @@ export function createNotificationStore() {
 }
 
 export const notificationsStore = createNotificationStore()
+
+subscribeBroadcast('notifications:missedCall', ({ token, name, avatar }) => {
+	const notification = new Notification(t('talk_desktop', 'You have missed a call from {name}', { name }), {
```
Antreesy, line 383: Align with what Talk backend sends? Would be easier to translate (or even copy-paste today from spreed repo)
```suggestion
	const notification = new Notification(t('talk_desktop', 'You missed a call from {user}' { user: name }), {
```

src/callbox/renderer/CallboxApp.vue
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Dismiss the call
+ */
+function dismiss() {
+	window.close()
+}
+</script>
+
+<template>
+	<div class="callbox" @keydown.esc="dismiss">
```
Antreesy, line 52: `@keydown.esc` doesn't work for me (guess that's because div is not a focused element, so `useHotKey()` should be used?)

src/callbox/renderer/CallboxApp.vue
```diff
@@ -0,0 +1,175 @@
...
+}
+</style>
+
+<style scoped>
+.callbox {
+	--spacing-1: 4px;
+	--spacing-2: 8px;
+	--spacing-3: 12px;
+	--spacing-4: 16px;
+	--golden-ratio: 1.618;
```
Antreesy, line 111: unused, left from first design attempt or needed?

Antreesy (approved):
Nothing blocking from me, tested in a group call

src/callbox/renderer/callbox.service.ts
```diff
@@ -0,0 +1,86 @@
...
+		}[]
+	}
+}
+
+/**
+ * Get participants of a call in a conversation
+ * @param token - Conversation token
+ */
+async function getCallParticipants(token: string) {
+	const response = await axios.get<CallGetParticipantsForCallResponse>(generateOcsUrl('apps/spreed/api/v4/call/{token}', { token }))
```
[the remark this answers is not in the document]
nickvergessen: Let use this for now as it's what the clients also use. We could also do a specific cheapest endpoint in the future

src/callbox/renderer/CallboxApp.vue
nickvergessen: we should not have keyboard shortcuts if this window is automatically focused.
And if we have shortcuts they need to ignore the first second or something to make sure the user presses them intentionally in this window/dialog

nickvergessen: But we can address this in the future

src/talk/renderer/notifications/notifications.store.js
nickvergessen: > You have missed a call from {name}

Does not work well in other languages. The notification should also give you the conversation type, and then we can toggle between:
- https://github.com/nextcloud/spreed/blob/main/lib/Notification/Notifier.php#L1021 and
- https://github.com/nextcloud/spreed/blob/main/lib/Notification/Notifier.php#L1054

src/callbox/renderer/callbox.service.ts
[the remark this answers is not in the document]
nickvergessen: Actually we have a much cheaper API already in the notifications app:
https://github.com/nextcloud/notifications/blob/master/openapi.json#L742-L748

Simply check if the ID is still there

src/callbox/renderer/CallboxApp.vue
nickvergessen: > It is not, as noted in the description and was discussed. Grabbing focus is not acceptable.

Ok cool, then proceed.

PR #891, opened 2024-11-13
fix(config): config path ignores actual app path

ShGKme:
### ☑️ Resolves

* Config path ignores the app name:
  * Production and Dev instances share the same config file
  * Multi-account instances share the same config
* Config is supposed to app's data directory, which depends on the name (executable)
* However, it didn't work:
  * Constant with the path was defined on the module-level
  * Module is executed on `require`/`import` before `app.setName()` was called and name was set
* Fix: get the name only on config load, not on module import

### DIFF
--- a/src/app/AppConfig.ts
+++ b/src/app/AppConfig.ts
@@ -10,10 +10,21 @@ import { isLinux, isMac } from '../shared/os.utils.js'
 
 const APP_CONFIG_FILE_NAME = 'config.json'
 
-// Windows: C:\Users\<username>\AppData\Roaming\Nextcloud Talk\config.json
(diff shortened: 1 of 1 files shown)


src/app/AppConfig.ts
```diff
@@ -120,6 +131,7 @@ async function writeAppConfigFile(config: Partial<AppConfig>) {
  * Load the application config into the application memory
  */
 export async function loadAppConfig() {
+	console.log(getAppConfigFilePath())
```
Antreesy, line 134: Remove?

Antreesy (approved):

PR #893, opened 2024-11-13
fix: zoom from config is not applied

ShGKme:
### ☑️ Resolves

* Zoom sometimes is not applied
* Electron's issue

## How to test

1. Open talk (in dev mode is easier)
2. Change zoom to 120% by wheel/shortcut/settings
3. Open the config file manually
4. Change `zoomFactor` to 0.8
5. Reload the window (type `rs` to the console with `npm run dev` running)

### Before

The window is small, but the zoom is big and unchanged

### After

New zoom is applied correctly


src/welcome/welcome.window.js
```diff
@@ -26,17 +28,26 @@ function createWelcomeWindow() {
...
+	// Hide traffic light buttons on Mac
 	if (isMac()) {
-		// Hide traffic light buttons on Mac
 		window.setWindowButtonVisibility(false)
 	}
 
 	window.loadURL(WELCOME_WINDOW_WEBPACK_ENTRY)
 
+	// Setting zoomFactor with webPreferences is not enough.
+	// If is only applied if it was set after the window is ready.
```
nickvergessen, line 44:
```suggestion
	// It is only applied if it was set after the window is ready.
```

src/welcome/welcome.window.js
```diff
@@ -26,17 +28,26 @@ function createWelcomeWindow() {
...
+	// Affects all the following windows.
```
nickvergessen, line 45: This sounds weird

ShGKme: The electron's bug or this text specifically?

nickvergessen: the last line is not a real sentence :see_no_evil:

nickvergessen (approved):

Antreesy:
Titlebar is not adjsuted when first load
image

ShGKme:
> * open instance

Instance of the same app or a separate instance with a new name?

ShGKme:
> copied directory, so app name is the same?

Why did you do that? ༼ つ ◕_◕ ༽つ

But I'll check.

Issue #885, opened 2024-11-13
Once closed, app window opens no more by activation in Dock


i2h3:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Launch app
2. Close window
3. Interact with other apps or do whatever
4. Click on the dock icon to bring Nextcloud Talk back into foreground

### Expected behaviour
An app window is opened. This is the conventional behavior in the platform. There is existing window to reveal again. Also, there is no menu item to create a new window explicitly.

### Actual behaviour
Nothing happens.

## Desktop client

**Talk Desktop client version:** v1.0.0-rc.1

**Operating system:** macOS

**Operating system version:** Sequoia 15.1

**Microphone available:** yes

**Camera available:** yes

i2h3:
As a workaround, it is possible to click on the menu bar extra to the upper right. That immediately reveals the (apparently only hidden) window. On the other hand, the "Open" menu item in there is kind of redundant then.

PR #894, opened 2024-11-14
fix(mac): show window on activate when minimized to tray

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/885
* Similar to https://github.com/nextcloud/talk-desktop/pull/852
  * That PR only fixed it for closed windows like the login window
  * This PR also fixes it when the window is not closed but minimized to the system tray

---

Additional changes after discussions are forked to https://github.com/nextcloud/talk-desktop/pull/911

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -5,7 +5,7 @@
 
 const path = require('node:path')
 const { spawn } = require('node:child_process')
-const { app, dialog, BrowserWindow, ipcMain, desktopCapturer, systemPreferences, shell } = require('electron')
+const { app, dialog, ipcMain, desktopCapturer, systemPreferences, shell } = require('electron')
 const { setupMenu } = require('./app/app.menu.js')
 const { setupReleaseNotificationScheduler } = require('./app/githubReleaseNotification.service.js')
 const { enableWebRequestInterceptor, disableWebRequestInterceptor } = require('./app/webRequestInterceptor.js')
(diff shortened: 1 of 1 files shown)


i2h3:
> Only do that with the main Talk window. Close login window = quit the app

This one. I am used to that behavior from many apps. There is no point in keeping it running in the background when you are not signed in. With its own apps, Apple is more and more quitting apps entirely on their last window close, too. Obviously that is inappropriate for a chat app.

But I would not consider it to be _that_ important difference.

i2h3 (approved):

ShGKme:
@Ivansss Despite our discussion yesterday but having @i2h3's comment and @SystemKeeper reaction, I went this way.
- Talk window - reopen on the Dock icon click (never closed, lives in the system tray)
- Login/Welcome windows - quit app on window close

Is it fine, or you'd insist on reopen for the login window dock icon click?

ShGKme:
Actually, after a new discussion with designers, I'm not sure we should close the login window 😣

I'll separate this PR into 2 parts:
- Fixing reopening the Talk window
- Quitting from the login window

ShGKme:
PR is updated:
- This PR only fixes the issue
- The new behavior is moved to https://github.com/nextcloud/talk-desktop/pull/911

Antreesy (commented):
LGTM

PR #896, opened 2024-11-15
build(mac): default to universal arch and adjust scripts and docs

ShGKme:
### 🚧 Tasks

See by commits

- [x] Lint `package.json` and `scripts/prepare-release-packages.mjs`
- [x] Remove `build:all` and `package:all`
  - General recommendation is to always build binaries on the same platform
  - It is still possible
- [x] Default `release:package` to the current platform
  - Now on each platform you can just `npm run release:package` instead of `npm run release:package -- --linux`
- [x] **Default macOS build to `universal` architecture**
  - Installer is 56% bigger (140MB -> 219MB) but no confusion for not advanced users
  - Add separate scripts to build a specific arch if needed, or we change the decision later

All the docs are updated as well


ShGKme:
Force pushed a small typo fix in the docs

nickvergessen (approved):

PR #906, opened 2024-11-18
build: unify distribution file names

ShGKme:
### ☑️ Resolves

Currently, distribution (installer) file names are different from different packaging tools.
This PR unifies it in a format:
```
Nextcloud.Talk-<PLATFORM>-<ARCH>.ext
```

Where platform is renamed to a user-friendly name:
- `darwin` => `macos`
- `win32` => `windows`

While it is cool to have a version in the file name, it makes it hard to have a constant link to the installer. The version is removed.

Examples:
- `Nextcloud.Talk-macos-universal.dmg`
- `Nextcloud.Talk-windows-x64.exe`
- `Nextcloud.Talk-windows-x64.msi`
- `Nextcloud.Talk-linux-x64.flatpak`
- `Nextcloud.Talk-linux-x64.zip`

### Implementation notes

All packagers have different output paths, incl. folder structure, and filenames.
In most of them, there is no option to specify or override it 😒

Hotfix: manually move files to the correct location.

Alternatives
- create our own makers extending the original (doesn't always help)
- fork `@electron/forge` makers, add features and upstream
- Do not use electron-forge makers for distribution - use original tools directly

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -18,7 +18,7 @@ require('dotenv').config()
 const CONFIG = {
 	// General
 	applicationName: packageJSON.productName,
-	applicationNameSanitized: packageJSON.productName.replaceAll(' ', '-'),
+	applicationNameSanitized: packageJSON.productName.replaceAll(' ', '.'),
 	companyName: 'Nextcloud GmbH',
 	description: packageJSON.description,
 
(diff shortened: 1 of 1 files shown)


forge.config.js
```diff
@@ -18,7 +18,7 @@ require('dotenv').config()
 const CONFIG = {
 	// General
 	applicationName: packageJSON.productName,
-	applicationNameSanitized: packageJSON.productName.replaceAll(' ', '-'),
+	applicationNameSanitized: packageJSON.productName.replaceAll(' ', '.'),
```
nickvergessen, line 21: dash is a problem? Would prefer that

nickvergessen (approved):

forge.config.js
ShGKme: No. But I didn't like that space replacer and name parts separator are the same symbol 🐧

PR #916, opened 2024-11-21
fix: apply selected theme also to native parts of the app

ShGKme:
### ☑️ Resolves

A minor change in theming. Simplifying pages setup and respects selected theme in all parts of the app.

* Before we used Nextcloud theme via `[data-theme-*]`. It has 2 drawbacks:
  * Requires setting attrs initially on every window
  * Native parts of the app have a system theme
* Now using a native theme:
  * Requires settings only once on app init
  * Works on all the parts

### 🖼️ Screenshots

Example - a Light theme on a Dark system

🏚️ Before | 🏡 After
---|---
image | image


Antreesy (approved):
Tested

PR #922, opened 2024-11-22
build(windows): add Squirrel MSI

ShGKme:
### ☑️ Resolves

- Adds MSI installer with Deployment Tool for administrative environment (when the app is installed using Group Policies)
- Note: the installation is still per-user in the `%User%/AppData/Local`, not `C:/Program Files/`
- Alternative to the classic MSI installer

Details: https://github.com/Squirrel/Squirrel.Windows/blob/develop/docs/using/machine-wide-installs.md

### DIFF
--- a/forge.config.js
+++ b/forge.config.js
@@ -71,7 +71,8 @@ function fixArtifactName(artifactPath, platform, arch) {
 	const artifactName = path.basename(artifactPath)
 	const ext = path.extname(artifactName)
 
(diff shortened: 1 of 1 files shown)


ShGKme:
Doesn't work after signing...

Antreesy (approved):
LGTM

ShGKme:
Also tested by @Antreesy

PR #929, opened 2024-11-25
build(dev): make NEXTCLOUD_DEV_SERVER_HOSTS optional

ShGKme:
### ☑️ Resolves

* Make `NEXTCLOUD_DEV_SERVER_HOSTS` which sets hosts for CSP in the dev server **optional**
  * Allows starting development without touching ENV variables or `.env` file
  * No more `.env` required during development, unless non-default values needed
* Adjust `.env.example`
* Update "Development Setup" in the `README` according to the new change


Antreesy (approved):
LGTM, tested

PR #935, opened 2024-11-29
fix(help): improve version description in report

ShGKme:
### ☑️ Resolves

* `git describe` might be confusing:
  * `g` prefix in `gHASH` specifying "git SCM" might be confusing for those who are not familiar with `git describe`
  * Version from the last tagged commit doesn't include the currently developed version, for example, it shows `20.0.0-beta.3` for `main`, even when there is `20.1.0-rc.2` on the stable branch and `21.0.0-dev.0` is currently being developed
  * The number of commits since the last branch is useless (0 for non-detached, and kind of a random value for detached)
  * `head/` is not useful for branches

Display the version for tagged commits.

Display `version-hash (branch)` for untagged.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/webpack.renderer.config.js
+++ b/webpack.renderer.config.js
@@ -58,19 +58,30 @@ function createPatcherAliases(packageName) {
 }
 
 /**
- * Get the version tag via "git describe".
- * @example "v1.0.1" on a directly tagged commit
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
Looks more accurate now

Issue #927, opened 2024-11-25
Failed to load image: JavaScript Error in Nextcloud Talk Client


TheMatzy:
## Steps to reproduce
1. Launch the Nextcloud Talk Client.
2. Use the app normally (the error occurs randomly, specific steps are unclear).

## Installation
Installed via MSI for Windows 10 with the newest Version v1.0.0-rc.2

### Expected behaviour
The client should load all resources properly without triggering a JavaScript exception.

### Actual behaviour
The app randomly throws the mentioned error, making it unreliable for consistent usage. The error appears to be related to a missing icon file. The file path suggests icon.ico in the resources folder might be missing or inaccessible?
![Image](https://github.com/user-attachments/assets/dc7176ba-c18d-4cf2-9895-da2d7df98df8)

## Diagnosis and logs

<details>
<summary>Diagnosis report</summary>

Insert a diagnosis report here.

### Diagnosis report

| **Nextcloud Talk Desktop**      |                                                                                |
| ------------------------------: | ------------------------------------------------------------------------------ |
| **Version**                     | v1.0.0-rc.2                                                                    |
| **Built-in Talk version**       | v20.1.0-rc.2                                                                   |
| **Operating system**            | Windows_NT 10.0.19045 (Windows 10 Pro)                                         |
| **Executable Path**             | C:\Users\butschek\AppData\Local\NextcloudTalk\app-1.0.0-rc2\Nextcloud Talk.exe |
| ****Connected to****            | -                                                                              |
| **Nextcloud server version**    | 30.0.2                                                                         |
| **Nextcloud Talk version**      | 20.0.2                                                                         |
| **`notifications` app enabled** | ✅ yes                                                                          |
| **`notify_push` app enabled**   | ✅ yes                                                                          |

#### Application config

```json
{
  "theme": "default",
  "systemTitleBar": false,
  "monochromeTrayIcon": false,
  "zoomFactor": 1,
  "playSoundChat": "respect-dnd",
  "playSoundCall": "respect-dnd"
}
```

</details>

RTechSn:
I get this error every time I connect via RDP to a machine with Talk Desktop running.

Issue #878, opened 2024-11-12
JavaScript error when switching Theme (dark/light/automatic) on MacOS


BJKle:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Install latest client on MacOS
2. Start application
3. Switch theme from dark to light or automatic

### Expected behaviour
Switch Theme and change icon color

### Actual behaviour
A JavaScript Error
Icon color is black (in both appearances)

![Image](https://github.com/user-attachments/assets/96113061-d421-4d81-9b4f-d5be7004c250)
![Image](https://github.com/user-attachments/assets/739aa457-b0fe-4c75-9084-194df96b8d84)
![Image](https://github.com/user-attachments/assets/9b3a0147-48ca-4eb1-aaf3-06472ef3f33c)

## Desktop client

**Talk Desktop client version:** 1.0.0-rc.1

**Operating system:** MacOS

**Operating system version:** Sequoia 15.1

**Microphone available:** yes

**Camera available:** yes

## Server

**Nextcloud version:** Nextcloud Hub 8 (29.0.8)

**Talk app version:** 19.0.11

**Custom Signaling server configured:** no

**Custom TURN server configured:** no

**Custom STUN server configured:** no

### Client log

<details>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

ShGKme:
Update: on Linux error is the same.

![Image](https://github.com/user-attachments/assets/8acd14a9-053a-4ece-b387-2841c2b8fa98)


On Windows it is different:

![Image](https://github.com/user-attachments/assets/ad640db9-0b58-4c2d-b6c1-4c530336b89e)

ShGKme:
Update: only reproducible in a production build. Dev works fine...

PR #936, opened 2024-11-29
fix: temporary remove buggy theme change handling for tray icon

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/927
* Fix: https://github.com/nextcloud/talk-desktop/issues/878

For unknown reasons:
- `NativeTheme.on('updated')` is triggered without theme change, for example, on external monitor connected/disconnected, or background image is changed on Windows
- When it happens, it says that the tray is destroyed (even if it works afterwards)
- And it cannot find an image for the icon (even if it worked on init and if you explicitly trigger a change in settings)

Also, this is only reproducible with the production build (`npm run build`).

This feature is very nit. But the bug is very annoying.

Until we found a solution - remove this feature to avoid errors.

To test (on non-macOS):
1. Make a production build `npm run build`
2. Run. Just in case, select the monochrome tray icon.
3. Change system theme

### DIFF
--- a/src/talk/talk.window.js
+++ b/src/talk/talk.window.js
@@ -3,14 +3,13 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-const { BrowserWindow, nativeTheme } = require('electron')
+const { BrowserWindow } = require('electron')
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
Lint is not happy

Issue #975, opened 2024-12-09
Nextcloud talk window does not start after second app startup on Debian


edoudabear:
## Steps to reproduce
1. Install nextcloud flatpak on debian 12.7
2. Run the app, closing the window, then opening it again

### Expected behaviour & Actual behaviour
The window should appear, but instead it remains stuck on the loading window (with the clouds moving)

## Diagnosis and logs

<details>
<summary>Diagnosis report</summary>

### Diagnosis report

| **Nextcloud Talk Desktop**      |                                                                             |
| ------------------------------: | --------------------------------------------------------------------------- |
| **Version**                     | v1.0.0                                                                      |
| **Built-in Talk version**       | v20.1.0                                                                     |
| **Operating system**            | Linux 6.1.0-28-amd64 (#1 SMP PREEMPT_DYNAMIC Debian 6.1.119-1 (2024-11-22)) |
| **Executable Path**             | /app/lib/com.nextcloud.talk/Nextcloud Talk                                  |
| **Using Wayland**               | ✅ yes                                                                       |
| ****Connected to****            | -                                                                           |
| **Nextcloud server version**    | 30.0.2                                                                      |
| **Nextcloud Talk version**      | 20.1.0                                                                      |
| **`notifications` app enabled** | ✅ yes                                                                       |
| **`notify_push` app enabled**   | ❌ no                                                                        |

#### Application config

```json
{
  "theme": "default",
  "systemTitleBar": true,
  "monochromeTrayIcon": false,
  "zoomFactor": 1,
  "playSoundChat": "respect-dnd",
  "playSoundCall": "respect-dnd",
  "enableCallbox": "respect-dnd"
}
```

</details>

<details>
<summary>Client logs</summary>
No relevant error logs

</details>

PR #993, opened 2024-12-18
fix(linux): correctly detect second instance

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/975
* Regression from https://github.com/nextcloud/talk-desktop/pull/859

I forgot that on Linux `argv[0]` can be relative...


Antreesy (approved):
LGTM, but can't test

PR #1011, opened 2024-12-28
chore: get rid of `undici`

ShGKme:
### ☑️ Resolves

- `undici` is not needed as `fetch` is available in Node 20
- There was a problem with `undici` which required us to use a fixed version 
- Migrated to `Fetch API`


Antreesy (approved):
Tested

PR #1012, opened 2024-12-30
fix: add logout button to welcome screen

Antreesy:
### ☑️ Resolves

- allows to re-auth in case Network Error persists and client stuck in the infinite loop

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
2024-12-30_12h26_47 | 2024-12-30_12h26_12

### 🚧 Tasks

- [ ] ...

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -313,7 +313,7 @@ app.whenReady().then(async () => {
 	})
 
 	ipcMain.handle('authentication:logout', async (event) => {
-		if (createMainWindow === createTalkWindow) {
+		if (createMainWindow === createTalkWindow || createMainWindow === createWelcomeWindow) {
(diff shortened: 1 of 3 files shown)


nickvergessen:
> in case Network Error persists and client stuck in the infinite loop

We should maybe add that info to the screen, instead of looping forever?

Antreesy:
We don't use translated strings or Vue components on welcome screen, so if we won't keep it 'always English' for simplicity, that would require more efforts

nickvergessen:
Well I mean:
- No one plans to logout
- The user is not aware that the request failed?
- Should show a dedicated message for maintenance mode, talk being disabled, and other most common problems
- Should show a generic error for others with an "error code" (http status code) so the user can be helped

Issue #951, opened 2024-12-03
Screen share on Mac mutes microphone


vaneslie:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Start a call and share your screen, doesn't matter if it's the whole screen or a window.
2. The microphone gets muted instantly and it's impossible to unmute (no microphone available).
3. You have to leave the call and join again to activate your microphone again.

### Expected behaviour
Microphone shouldn't be muted during screen sharing.

### Actual behaviour

![Image](https://github.com/user-attachments/assets/f121e7f6-7bdc-4c53-965d-f732251e8084)

## Diagnosis and logs

<details>
<summary>Diagnosis report</summary>

Nextcloud Talk Desktop |  
-- | --
Version | v1.0.0-rc.2
Built-in Talk version | v20.1.0-rc.2
Operating system | Darwin 24.1.0 (Darwin Kernel Version 24.1.0: Thu Oct 10 21:05:14 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T8103)
Executable Path | /Applications/Nextcloud Talk.app/Contents/MacOS/Nextcloud Talk
Connected to | -
Nextcloud server version | 30.0.3 RC1
Nextcloud Talk version | 20.1.0-rc.3
notifications app enabled | ✅ yes
notify_push app enabled | ✅ yes

</details>

<details>
<summary>Client logs</summary>

Press <kbd>Ctrl+Shift+I</kbd> (<kbd>Cmd+Option+C</kbd> on macOS) or use the context menu to open the developer tools, then go to the Console tab.

```
Console tab is not clickable for me.
```

</details>

PierreLga456:
Same for me

nemphys:
Same here. This is definitely new, it started happenning after the latest update.

PR #1023, opened 2025-01-09
fix(macOS): microphone crash on screen sharing

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/951

Requesting an entire desktop with a stream with audio is not actually supported in macOS. Somehow an attempt to request it results in losing audio devices access with

```
The AudioContext encountered an error from the audio device or the WebAudio renderer.
```

Current workaround is as simple as removing an unsupported option.

An alternative is to use experimental native screen sharing selector. But it is more complex, requires changes on the Talk side and loses an Entire desktop option for Windows and Linux.
- https://github.com/nextcloud/talk-desktop/pull/1022

## Screenshots

See mic icon. Before it is off and disabled due to an error.

Before | After
---|---
Screenshot 2025-01-15 at 16 21 51 | Screenshot 2025-01-15 at 16 19 01


Antreesy (approved):
Sounds good for a quick solution

PR #1029, opened 2025-01-15
build: add release channels

ShGKme:
### ☑️ Resolves

- Allows to build/release Talk for different release channels:
  - `stable` - on stable Talk
  - `beta` - on beta or rc Talk
  - `dev` - on main (or any other commit)
- Versions are stored in `package.json`
- Managed by `CHANNEL` env variable and `--channel` arg in the release script
- Non-stable release is marked on UI

### 🖼️ Screenshots

Non-stable channel in the title bar and diagnosis report

image

On macOS:

Screenshot 2025-01-19 at 22 12 33

Full version on the Login window on non-stable 

image

Non-stable welcome window

image

### 🚧 Tasks

- [ ] ...

### DIFF
--- a/src/talk/renderer/TitleBar/TitleBar.vue
+++ b/src/talk/renderer/TitleBar/TitleBar.vue
@@ -43,17 +43,18 @@ useHotKey('Escape', pushToRoot)
 <template>
 	<header id="header" class="title-bar">
 		<div class="title-bar__inner">
-			<div v-if="!OS.isMac"
-				class="title-bar__title"
-				role="button"
-				tabindex="0"
-				@click="pushToRoot">
-				Nextcloud Talk
-			</div>
+			<template v-if="!OS.isMac">
+				<div class="title-bar__title"
+					role="button"
+					tabindex="0"
+					@click="pushToRoot">
+					Nextcloud Talk
+				</div>
 
-			<div v-if="channel !== 'stable'" class="title-bar__channel">
-				{{ channel }}
-			</div>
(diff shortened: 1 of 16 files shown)


.env.example
```diff
@@ -8,3 +8,9 @@ TALK_PATH=spreed
 # Defaults to "*" which might be unsafe during development - you can set it to allow only your hosts
 # Example: https://nextcloud.local wss://signaling.nextcloud.local wss://notify_push.nextcloud.local
 NEXTCLOUD_DEV_SERVER_HOSTS=*
+
+# Application channel - stable or preview
```
nickvergessen, line 12: would using "beta" be problematic? Then the naming would match the one on server

nickvergessen (approved):
Otherwise good

src/talk/renderer/TitleBar/TitleBar.vue
```diff
@@ -44,12 +44,15 @@ useHotKey('Escape', pushToRoot)
...
 				role="button"
 				tabindex="0"
 				@click="pushToRoot">
-				<span class="title-bar__title">Nextcloud Talk</span>
-				<span v-if="isPreview" class="title-bar__preview-badge">Preview</span>
+				Nextcloud Talk
+			</div>
+
+			<div v-if="channel !== 'stable'" class="title-bar__channel">
+				{{ channel }}
```
nickvergessen, line 55: Let's only do that for non-MacOS and add the channel name to the menu bar of macOS as well

Antreesy (approved):
Code-wise looks good, (only tested a Windows build, visually - OK)

PR #1032, opened 2025-01-16
refactor: TitleBar

ShGKme:
## ☑️ Resolves

* Preparation for initialization refactoring, to fix `@nextcloud/vue` upgrade regression

A huge change, but no logic changed.

Separated into small commits - review by commits.

### 1. Rename `DesktopHeader` to `TitleBar`

It was called `header` because it represented Nextcloud web-page's `#header` in the beginning.

But in the app, it represents desktop app's title bar.

### 2. Move files to a new module folder

To keep coupled modules in one place.

### 3. Migrate UserStatus to TS

Helped with the next step.

Also fix some typing issues, texts, and formatting.

### 4. Migrate TitleBar to TS

Ts.

### DIFF
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -28,6 +28,21 @@ declare module '@nextcloud/vue/dist/Components/*.js' {
 	const component: Component
 	export default component
 }
+declare module '@nextcloud/vue/dist/Composables/useHotKey.js' {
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
Tested. Looks clean!

Never have seen `as const` though 👀

PR #1033, opened 2025-01-17
fix: BiDi support

ShGKme:
### ☑️ Resolves

* Add BiDi support, including:
  * Setting `dir` based on language
  * Correctly handle RTL in title bar (CSS env if always LTR X position)

Note for reviewers: since Talk Desktop only uses system language, for testing you can change it in

https://github.com/nextcloud/talk-desktop/blob/7227939b7b505d2d113e192d40433b22add01129/src/shared/setupWebPage.js#L57

### 🖼️ Screenshots

image


DorraJaouad (approved):
Looks good

Issue #389, opened 2023-10-27
Desktop client is reloading while being in a call


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Yesterday I had a call with a coworker
2. I started updating our instance
3. After the update I saw the notification in the top right that it needed a reload

### Expected behaviour
After ending the call the next click on "start call"/"join call" or another conversation reloads, just like in the web

### Actual behaviour
It reloaded while I was still in the call, interrupting the experience.

## Desktop client

```
----------------------------System report----------------------------
Nextcloud Talk Desktop version 0.12.0
- Built with Nextcloud Talk version 17.1.1

Connected to:
- Server address: https://cloud.nextcloud.com
- Nextcloud Server version 27.1.3
- Nextcloud Talk version 17.1.1

OS: Linux 6.5.0-10-generic (#10-Ubuntu SMP PREEMPT_DYNAMIC Fri Oct 13 13:49:38 UTC 2023)
----------------------------System report----------------------------
```

## Server

**Nextcloud version:** 27.1.2, now 27.1.3

**Talk app version:** 17.1.1

nickvergessen:
> When I click "join a call" I want to either join it or see any message that explains what I haven't joined the call. The same with the conversations list.

In the browser the "join/start call" button is disabled when this is a case and telling you it needs a reload:
https://github.com/nextcloud/spreed/blob/b9c0f8318519ce7421b76c42d3d69927d1230566/src/components/TopBar/CallButton.vue#L235-L242

Well basically everything but staying in the current state can be an issue. So when deleting a message, writing a message, making a reaction or something alike fails we should prompt for the pending reload. If they still work, fine by me.
Switching a conversation and joining a call have to actively be intervened because the dirty mark could be that turn, stun, hpb, recording or other server settings where changed and you need to use another server for the calls, etc.

nickvergessen:
This happened to me multiple times yesterday when configuring the new beta features on our cloud in a coworking call.

PR #1050, opened 2025-01-22
fix: do not immediately relaunch app on Talk server update

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/389

#### Before

Immediately relaunch the app on dirty hash!

#### In the Web

A user can continue use the app, but:
- There is a toast
- Starting a call is not available
- Opening a new chat reloads the page

#### After

Handled almost by Web app part:
- There is a toast
- Starting a call is not available
- Instead of opening a new chat link in a web-browser or as a web-page inside an app:
  - Navigate to the new route internally
  - Reload a page
  - Having dirty hash - refetch capabilities before initializing the app

There is one potential problematic scenario:
- Page is reload
- But the server and/or internet connection is extremely slow
- Where is no indicator that something is loading. Just a blue window

### 🖼️ Screenshots

image

https://github.com/user-attachments/assets/100ba089-62d2-45b7-b629-10688f7ba93d

### DIFF
--- a/src/shared/setupWebPage.js
+++ b/src/shared/setupWebPage.js
@@ -11,6 +11,7 @@ import { initGlobals } from './globals/globals.js'
 import { setupInitialState } from './initialState.service.js'
 import { initAppConfig } from './appConfig.service.ts' // eslint-disable-line import/namespace
 import { TITLE_BAR_HEIGHT } from '../constants.js'
+import { refetchAppData } from '../app/appData.service.js'
 
 /**
  * @param {string} lang - language code, TS type: `${lang}_${countryCode}`|`${lang}`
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
Video looks good :)

src/app/externalLinkHandlers.ts
```diff
@@ -30,6 +31,32 @@ export function isExternalLink(url: string) {
...
+	if (!appData.serverUrl || url.origin !== appData.serverUrl) {
+		return null
+	}
+
+	// Path without /index.php
+	const pathname = url.pathname.slice(url.pathname.startsWith('/index.php') ? '/index.php'.length : 0)
+
+	// Check whether the path is a valid Talk path or some other app
+	// eslint-disable-next-line prefer-regex-literals
+	const talkPathRe = new RegExp('^/apps/spreed(?:/(?:not-found|forbidden|duplicate-session))?|/call/[^/]+$')
```
Antreesy, line 51: Is there a specific reason to not define it:
- outside of function?
- via constructor?

Antreesy (approved):
Code-wise also looks good

PR #1052, opened 2025-01-24
chore: update Build-in Talk script and CI

ShGKme:
image

### DIFF
--- a/.github/workflows/update-built-in-talk.yml
+++ b/.github/workflows/update-built-in-talk.yml
@@ -1,0 +1,51 @@
+# SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: CC0-1.0
+
+name: Update Built-in Talk
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/update-built-in-talk.yml
```diff
@@ -0,0 +1,51 @@
...
+          # Pass the result to the next steps
+          echo "stdout<<EOF" >> $GITHUB_OUTPUT
+          echo "$stdout" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+      - name: Print result
+        run: echo "${{ steps.update.outputs.stdout }}"
+
+      - name: Create Pull Request
+        if: steps.update.outcomes == 'success'
```
SystemKeeper, line 40:
```suggestion
        if: steps.update.outcome == 'success'
```

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#steps-context

SystemKeeper: Also is there any error case? So if no changes are found, it would still create a PR? (Just asking)

SystemKeeper (changes requested):

PR #1073, opened 2025-01-30
fix: remove unneeded capabilities fetching on every window open

ShGKme:
### ☑️ Resolves

* Issue: window init is always long
* Happens after https://github.com/nextcloud/talk-desktop/pull/1050

1. Pinia executes `setNextcloudTalkHash` even when there is the same hash already (check is inside)
2. Talk Desktop is sure, `setNextcloudTalkHash` is only dispatched when there is a NEW hash, so it marks the hash as dirty...
3. Each second load is marked as dirty and fetches capabilities...

Either I missed a difference between new Pinia and old Vuex hash store implementation. Or I'm silly and there always was an issue 🤦

- Fixed the issue
- Renamed `talkHashSetInitial` to `talkHashUpdate` to not confuse

### DIFF
--- a/src/talk/renderer/TalkWrapper/TalkWrapper.vue
+++ b/src/talk/renderer/TalkWrapper/TalkWrapper.vue
@@ -5,7 +5,7 @@
 
 <script setup lang="ts">
 import { onMounted } from 'vue'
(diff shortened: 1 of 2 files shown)


Antreesy (approved):
Tested

PR #1083, opened 2025-02-03
fix(upgrade): adjust styles

ShGKme:
### ☑️ Resolves

* Slightly adjust styles:
  * Use Nextcloud grid baseline
  * Adjust spacing
* Migraste to TS

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


src/upgrade/renderer/UpgradeApp.vue
```diff
@@ -3,55 +3,92 @@
...
+				{{ t('talk_desktop', 'Update Talk Desktop ↗') }}
+			</NcButton>
+			<NcButton :href="browserLink"
+				type="secondary"
+				target="_blank"
+				wide>
+				<template #icon>
+					<IconWeb :size="20" />
+				</template>
+				{{ t('talk_desktop', 'Continue in Web-Browser ↗') }}
```
nickvergessen, line 45:
```suggestion
				{{ t('talk_desktop', 'Continue in Web-Browser') }} ↗
```

Saving translators some pain? Same above?

But now I see this was part before so it would be a new string now.

src/upgrade/renderer/UpgradeApp.vue
```diff
@@ -3,55 +3,92 @@
...
 
-.upgrade h2,
-.upgrade p {
-	user-select: none;
+.upgrade__content {
+	display: flex;
+	flex-direction: column;
+	justify-content: space-evenly;
+	gap: var(--upgrade-spacing);
+	flex: 1;
```
DorraJaouad, line 76: image
It doesn't look in its best shape when expanded , it would be better if they don't expand and get centered.

src/upgrade/renderer/UpgradeApp.vue
```diff
@@ -3,55 +3,92 @@
...
 </template>
 
+<style>
+* {
+	box-sizing: border-box;
+	user-select: none;
+}
+</style>
+
 <style scoped>
```
DorraJaouad, line 58: It's the same if it is put in `scoped` as it's the only instance mounted?

src/upgrade/renderer/UpgradeApp.vue
DorraJaouad: > Did you test in Windows Subsystem Linux?

Yes

Antreesy (commented):
Otherwise style changes are fine. But would be nice to see it on mac/ real linux

Antreesy:
> Could you show how flexbox looks for you on devtools?

Was an issue on my side, resolved now

Antreesy (approved):

PR #1103, opened 2025-02-13
build: windows .msi via WiX

ShGKme:
### Resolves

- We have .msi as a Squirrel Deployment Tool
- This Deployment Tools is installed machine-wise as expected
- But it is not the app itself. On the next user login, it installs Nextcloud Talk using the user-wise installer into `%AppData%/Local`
- It doesn't work for some administrated environments, expecting the app itself to be installed machine0wise into `%ProgramFiles%`
- This installer is a more classic .msi installer with WiX v3

### 🖼️ Screenshots

image

image

image

image

image


Antreesy (approved):
Tested

PR #1107, opened 2025-02-14
chore: remove moment

ShGKme:
### ☑️ Resolves

* `@nextcloud/moment`:
  * uses unmaintained `moment`
  * uses old `@nextcloud/l10n` v2
  * isn't really needed in modern Chromium

### DIFF
--- a/src/shared/datetime.utils.ts
+++ b/src/shared/datetime.utils.ts
@@ -54,10 +54,12 @@ export function formatRelativeTimeFromNow(dateOrMs: Date | number) {
 }
 
 /**
- * Convert milliseconds to the largest rounded unit
+ * Convert milliseconds to the largest unit rounded from 0.75 point.
  * @example 123 -> { value: 0, unit: 'second' }
  * @example 1000 -> { value: 1, unit: 'second' }
  * @example 25 * 60 * 60 * 1000 -> { value: 25, unit: 'minute' }
+ * @example 35 * 60 * 60 * 1000 -> { value: 35, unit: 'minute' }
+ * @example 45 * 60 * 60 * 1000 -> { value: 1, unit: 'hour' }
  * @example 3600000 -> { value: 1, unit: 'hour' }
  * @example 86400000 -> { value: 1, unit: 'day' }
  * @param ms - Duration in milliseconds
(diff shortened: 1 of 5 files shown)


src/talk/renderer/UserStatus/userStatus.utils.ts
```diff
@@ -4,8 +4,8 @@
  */
 
 import type { ClearAtPredefinedConfig, PredefinedUserStatus, UserStatus, UserStatusStatusType } from './userStatus.types.ts'
-import moment from '@nextcloud/moment'
-import { translate as t } from '@nextcloud/l10n'
+import { t, getFirstDay } from '@nextcloud/l10n'
```
Antreesy, line 7: Screenshot from `appData.userMetadata`, this results in empty `<html data-locale >` (when was not set on personal settings). Maybe it should defaults to selected language; here to 'de'?

image

also custom `window.firstDay` from settings is not provided as in web, and always leads to default locale value:
- 'en' - Sunday
- 'de' - Monday

Antreesy (approved):
Unrelated to changes comment. Otherwise works fine

src/talk/renderer/UserStatus/userStatus.utils.ts
ShGKme: > also custom `window.firstDay` from settings is not provided as in web

Yes, this info is not available for clients, so, using locale data

ShGKme: > Screenshot from `appData.userMetadata`, this results in empty `<html data-locale >` (when was not set on personal settings). Maybe it should defaults to selected language; here to 'de'?

I cannot reproduce this state... Even for new accounts it's `en` initially. How do you unset it?

Anyway, Nextcloud defaults it to `'en'` in `@nextcloud/l10n`, so should be fixed there. But I'd fallback for the local device settings, not the language.

Antreesy: `occ user:setting <userid> --delete core locale`. Then in web it defaults to `core.lang`

ShGKme: Proposal: https://github.com/nextcloud-libraries/nextcloud-l10n/pull/864

src/shared/datetime.utils.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { getCanonicalLocale } from '@nextcloud/l10n'
```
Antreesy, line 6: Late-thought opinion: `getLanguage` should be used for relative time (not absolute, that required locale format):

With 'en' language and 'de' locale we are geting human-readable represenation, but language of it won't match expected system:

Locale \ Lang | 'en' | 'de'
-- | -- | --
'en' | 1:23 PM, 'in 1 day' 🟢  | 13:23, 'in 1 day'
'de' | 1:23 PM, 'in 1 Tag' | 13:23, 'in 1 Tag' 🟢

ShGKme: Yes, and this is an old discussion about using locale for durations.

Here it works the same way it works on the server.

PR #1118, opened 2025-02-17
fix: mount apps from async components

### LINKED ISSUE #1117: Call notification (callbox) is still active when the call is joined already
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.

---

## Steps to reproduce
1. Have an incoming call and call notification
2. Join the call

### Expected behaviour

The callbox should disappear

### Actual behaviour

The callbox is still on the screen

Antreesy:
### ☑️ Resolves

* Fix #1117
* Ref #1013
  * same solution should be propagated on all apps - `nextcloud/*` imports should be async, to load and initialize after setting up webpage (e.g. globals and user metadata)


src/talk/renderer/TalkDesktop.app.ts
```diff
@@ -15,8 +15,11 @@ export async function createTalkDesktopApp() {
 
 	// Load Talk Desktop asynchronously to make sure,
 	// no module is loaded before the page has been set up and ready to run apps
-	const { default: TalkDesktop } = await import('./TalkDesktop.vue')
+	const TalkDesktopApp = defineAsyncComponent(() => import('./TalkDesktop.vue'))
```
ShGKme, line 18: There was async import already, works fine without an additional component wrapper.

Antreesy: I wanted to unify it between all windows, but if you say it's not needed, let's drop it

ShGKme: If unify, I'd drop it everywhere then: to have fewer wrappers in the component tree

ShGKme (approved):

Issue #1116, opened 2025-02-17
Repeat ringtone on call popup


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Have an incoming call notification with callbox enabled

### Expected behaviour

It rings all the time with a short interval.

### Actual behaviour

It rings only once for 5-10 seconds, and then it's silent for the remaining 35 seconds.

PR #1123, opened 2025-02-19
fix(callbox): loop ringtone

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/talk-desktop/issues/1116
* Added devmode trigger for the call box
* Moved call ringtone to the callbox
  * No ringtone with a simple notification anymore (as expected)
* Loop the ringtone


Antreesy (approved):
Tested.

Now it's obvious (and should also be the issue in notifications app):
image

If device is the same (but device id is different - see all entries of Headphone) - sound will be duped, causing the echo.

We need to find a way to detect it differently somehow

PR #1136, opened 2025-02-24
fix(user-status): emoji picker cannot be focused

ShGKme:
### ☑️ Resolves

* Hotfix for emoji picker focus trap
* Regression from https://github.com/nextcloud-libraries/nextcloud-vue/pull/6466
* A proper solution on the nextcloud-vue side is in progress

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before | after


Antreesy (approved):
Tested

PR #1152, opened 2025-03-04
fix(config): remove notification on dnd

ShGKme:
### ☑️ Resolves

* This option is not available in general case
* Removed for consistency across clients

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


nickvergessen (approved):
We could keep the non-bool values for the future, but fine either way

ShGKme:
Boolean values reflect corresponding config values on the server
image

Issue #985, opened 2024-12-17
Nextcloud Talk Windows Client fails to upload data from server (CSRF mismatch, 404 errors, and incorrect locale errors)


nassimshahin:
I am experiencing an issue with the Nextcloud Talk Windows Client where it fails to upload data from the Nextcloud server. The developer console shows multiple errors, including:

resources:
bash

404 Not Found errors when accessing user_status, dav/events/upcoming, and other resources:

CSRF token mismatches:
communis.erzbistum-koeln.de/ocs/v2.php/apps/user_status/api/v1/statuses/_<UUID>:1 
Failed to load resource: the server responded with a status of 404 (Not Found)

arduino
Request to https://communis.erzbistum-koeln.de/index.php/apps/files/api/v1/configs failed because of a CSRF mismatch. Fetching a new token

javascript
RangeError: Incorrect locale information provided
    at Intl.Collator (<anonymous>)
    at J.t (index.js)

javascript
Uncaught SyntaxError: Unexpected token 'E', "Error whil"... is not valid JSON
at JSON.parse

csharp
index.js:1 Session has been marked as inactive
index.js:1 Session has been marked as active

Nextcloud Talk Client
Server: Nextcloud server (URL: communis.erzbistum-koeln.de)
Language/Locale: de_DE

nassimshahin:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.


---

## Steps to reproduce

1. Access the **Nextcloud Talk** or related app through the provided URL (`communis.erzbistum-koeln.de`).
2. Attempt to join a call room (`/call/r2nkd9ed`).
3. Observe console errors and connection failures.

### Expected behaviour

The connection to the server should work seamlessly, and calls should be successfully initialized without errors.

### Actual behaviour

Several issues occur:
1. **404 Not Found** responses when requesting the following endpoints:
   - `.../ocs/v2.php/apps/user_status/api/v1/statuses/...`
   - `.../ocs/v2.php/apps/dav/api/v1/events/upcoming`
2. **CSRF mismatch errors** when accessing configurations:
   - `.../index.php/apps/files/api/v1/configs`
[...]

PR #1164, opened 2025-03-11
fix: substitute empty locale from server with system locale

Antreesy:
### ☑️ Resolves

* Issue #985
* appData.userMetadata might have a locale as `" "` empty string (which means user has never changed it in Personal settings), but it won't be replaced with system value and handled by system as empty string
  image
* `Intl.*` constructors do not accept empty string as a locale:
image

### 🖼️ Screenshots

image

### 🚧 Tasks

- [ ] ...


ShGKme (approved):
Thanks!

DorraJaouad (approved):

Issue #1171, opened 2025-03-15
Improve Application Badge Count Implementation


markusstoll:
This is a follow up to

https://github.com/nextcloud/talk-desktop/issues/85

## Current Situation & Problem
The current implementation of application badge counts in Talk Desktop is insufficient and unreliable, which has been a major hindrance for productive use of the desktop application. The main issues are:
- Badge don't update reliably when new messages (actually it almost never shows up and if it does it does not show a number of unread elements)
- Counts are sometimes incorrect after starting the application

This has led many users to avoid using the desktop application, falling back to browser-based solutions instead.

## Architectural Changes
The new implementation introduces a clean separation of concerns:
1. The badge count is now managed through a dedicated event-based system
2. The `TalkWrapper` component only:
   - Registers the unread count store
   - Listens to the `UNREAD_COUNT_UPDATED` event
   - Updates the badge count via the existing `setBadgeCount` IPC interface
3. All calculations are delegated to the Talk core (spreed) component

## Implementation Details
The changes in talk-desktop are minimal and focused:
1. Remove outdated badge count handling code
2. Utilize the existing `setBadgeCount` IPC interface
3. Add event listener for the new `UNREAD_COUNT_UPDATED` event from Talk core
4. Update the badge count when events are received

All of required changes are already done and work fine for handful of testers (tested with MacOs and Windows so far).
Due to the fact that I need changes in spreed AND talk-desktop, my further procedure is 

- create a similar issue in nextcloud/spreed
- create a pull request in nextcloud/spreed
- create a pull request here

PR #1174, opened 2025-03-17
Use new unreadCountStore for setting appIcon Badge

markusstoll:
### ☑️ Resolves

* Issue https://github.com/nextcloud/talk-desktop/issues/1171
* spreed ref https://github.com/nextcloud/spreed/pull/14668

All changes are described in the mentioned issue

### DIFF
--- a/src/talk/renderer/TalkWrapper/TalkWrapper.vue
+++ b/src/talk/renderer/TalkWrapper/TalkWrapper.vue
@@ -9,6 +9,7 @@ import { onTalkHashDirty, onTalkHashUpdate, openConversation, setTalkHash } from
 import { registerTalkDesktopSettingsSection } from '../Settings/index.ts'
 import { subscribeBroadcast } from '../../../shared/broadcast.service.ts'
 import { appData } from '../../../app/AppData.js'
+import { subscribe } from '@nextcloud/event-bus'
 
 const emit = defineEmits<{
 	(event: 'ready'): void
@@ -32,6 +33,10 @@ onMounted(async () => {
 	})
 	onTalkHashDirty(() => {
 		appData.setTalkHashDirty(true).persist()
+	})
+
+	subscribe('talk:unread:updated', ( eventData ) => {
+		window.TALK_DESKTOP.setBadgeCount(eventData.messages)
 	})
 
 	// Ready
--- a/src/talk/renderer/notifications/notifications.store.js
+++ b/src/talk/renderer/notifications/notifications.store.js
@@ -94,7 +94,6 @@ export function createNotificationStore() {
 		if (notifications.length > _oldcount) {
 			_oldcount = notifications.length
 			if (state.backgroundFetching && document.hidden) {
-				window.TALK_DESKTOP.setBadgeCount()
 				window.TALK_DESKTOP.flashAppIcon(true)
 				// If we didn't already highlight, store the title so we can restore on tab-view
 				if (!document.title.startsWith('* ')) {
@@ -110,7 +109,6 @@ export function createNotificationStore() {
 	 * the Talk might have altered it.
 	 */
 	function _restoreTitle() {
-		window.TALK_DESKTOP.setBadgeCount(0)
 		window.TALK_DESKTOP.flashAppIcon(false)
 		if (document.title.startsWith('* ')) {
 			document.title = document.title.substring(2)


src/talk/renderer/TalkWrapper/TalkWrapper.vue
```diff
@@ -4,38 +4,85 @@
...
-	// Additional integrations
-	registerTalkDesktopSettingsSection()
-	subscribeBroadcast('talk:conversation:open', ({ token, directCall }) => openConversation(token, { directCall }))
+		// Initialize badge handling
+		window.TALK_DESKTOP.setBadgeCount(0)
+
+		// Subscribe to unread count updates
+		unsubscribeUnreadCount = subscribe(EVENTS.UNREAD_COUNT_UPDATED, ({ unreadMessages }) => {
+			window.TALK_DESKTOP.setBadgeCount(unreadMessages || 0)
+		})
```
Antreesy, line 60: IMO this is the only part that belongs to Desktop client:
- it should be subscribed to talk event
- it should update counter based on internal logic and client config (say, user wants to ignore them at all and don't show a badge / or count only unread mentions / or all of unread messages)

Store should be properly registered in the app itself

Antreesy (commented):
Gave it a smoke test, seems to be working as expected 🎉

Leaving my thoughts for this PR. I suggest you to wait for another devs opinions, before rewriting anything; then we can find a common ground before putting more efforts into this feature.

src/talk/renderer/TalkWrapper/TalkWrapper.vue
markusstoll: Not that I want anyone to rush things, but I understood correctly that we shall wait for more input from your devs?

nickvergessen: @ShGKme wanted to post feedback soon

src/talk/renderer/TalkWrapper/TalkWrapper.vue
```diff
@@ -4,38 +4,85 @@
...
+		registerTalkDesktopSettingsSection()
+		subscribeBroadcast('talk:conversation:open', ({ token, directCall }) => openConversation(token, { directCall }))
+
+		await waitForTalkAndStore()
+		const store = window.store
+
+		// Register store module if not already registered
+		if (!store.hasModule('unreadCount')) {
+			store.registerModule('unreadCount', unreadCountStore)
+		}
```
ShGKme, line 52: I appreciate the willing to not register this module in `spreed` repo as this isn't used there, but it is barely safe anything performance wise and makes the solution much more complicated.

It adds an additional place where Talk Desktop depends on Talk Web internal implementation. Even migrating from Vuex to Pinia or another refactoring requires changes on the Talk Desktop side. It is especially a problem because we support building Talk Desktop with at least two Talk versions (for stable and beta releases).

This also requires importing from Talk Web source code. While we do it in several places already, this might be a problem in the future, and we should avoid it when possible by using public frontend API.

Instead, let's only rely on the global event via `subscribe`. This would also allow simplifying the coded added here.

ShGKme: Same for the `EVENTS.UNREAD_COUNT_UPDATED` constant. While it is more reliable to have the constant instead of a raw string here, it requires importing the constant from another repository.

Alternatives:
1. Use a raw string. TS can safe here from making a typo
2. Add an explicit (un)subscribing method to public API via `window.OCA.Talk`

ShGKme (changes requested):

ShGKme (changes requested):

src/talk/renderer/TalkWrapper/TalkWrapper.vue
```diff
@@ -34,6 +36,16 @@ onMounted(async () => {
...
 
+	subscribe(EVENTS.UNREAD_COUNT_UPDATED, (eventData) => {
+		// Process unread counter values
+		const { totalUnreadMessages, totalUnreadMentions, totalUnreadMentionsDirect } = eventData
+		
+		if (window.TALK_DESKTOP) {
+			const badgeCount = totalUnreadMessages
+			window.TALK_DESKTOP.setBadgeCount(badgeCount)
+		}
+	})
```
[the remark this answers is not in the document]
markusstoll: Simplified as discussed, but adapted to the actual implementation of the event in spreed - see PR there

markusstoll:
As with my changes in spreed I squashed all changes into 1 commit

PR #1181, opened 2025-03-26
chore: migrate to ESLint 9 and new @nextcloud/eslint-config

ShGKme:
### ☑️ Resolves

* Migrating from ESLint 8 with a deprecated config format
* Some rules are temporally disabled and are to be fixed in upstream


eslint.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+			},
+		},
+		settings: {
+			jsdoc: {
+				tagNamePreference: {
+					return: 'return', // TODO: upstream
+				},
+			},
+		},
+	},
```
ShGKme, line 52: To use `@return` instead of `@returns`

eslint.config.mjs
```diff
@@ -0,0 +1,131 @@
...
+			'@typescript-eslint/no-shadow': 'off', // TODO: upstream
+			'@typescript-eslint/no-use-before-define': ['error', { functions: false }], // TODO: upstream
+		},
+	},
+	{
+		files: ['**/*.vue'],
+		rules: {
+			'vue/require-prop-comment': 'off', // TODO: upstream
+		},
+	},
```
ShGKme, line 93: Only these rules are "new".

All the other parts in the file are the old config migrated to a new Flat config format

Antreesy (approved):

ShGKme:
@DorraJaouad If oyu wanted to review - you still can do it, and I'll make adjustments according to your comment if necessary :)

PR #1193, opened 2025-04-04
chore: build config

ShGKme:


build/BuildConfig.types.ts
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	domain: string | null
+
+	/**
+	 * Default: false
+	 */
+	enforceDomain: boolean
+
+	/**
+	 * Whether multi account is enabled.
```
Antreesy, line 50: Also NOT IMPLEMENTED?

build/resolveBuildConfig.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+const path = require('node:path')
+const fs = require('node:fs')
+const defultConfig = require('./config.json')
```
Antreesy, line 8: 👀

src/authentication/renderer/AuthenticationApp.vue
```diff
@@ -241,4 +252,9 @@ async function login() {
 .submit-button {
 	margin-top: 0.5rem;
 }
+
+:deep(.login-box__server--predefined) {
```
Antreesy, line 256: styles vary a bit, but that might be upstream issue
loading | error
-- | --
image | image

ShGKme: I didn't understand. Do you mean the font color?

build/BuildConfig.types.ts
ShGKme: Yes

build/resolveBuildConfig.js
Antreesy: It's defult

PR #1200, opened 2025-04-06
Add custom showCertificateTrustDialog for Linux

Captainpast:
### ☑️ Resolves

* Issue #23

### 🖼️ Screenshots

grafik

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -218,49 +218,69 @@ app.whenReady().then(async () => {
 		}
 	})
 
+	/** To store pending "trust certificate" dialogs on Linux */
+	const openMessageBoxes = new Map()
 	app.on('certificate-error', (event, webContents, url, error, certificate, callback) => {
 		event.preventDefault()
 
 		if (isLinux) {
-			return callback(process.env.NODE_ENV !== 'production')
+			const trustedFingerprints = getAppConfig('trustedFingerprints') ?? []
+			if (trustedFingerprints.includes(certificate.fingerprint)) {
+				callback(true)
+			} else {
+				let messageBox = null
+				if (openMessageBoxes.has(certificate.fingerprint)) {
+					messageBox = openMessageBoxes.get(certificate.fingerprint)
+				} else {
+					messageBox = dialog.showMessageBox(mainWindow, {
+						type: 'warning',
+						title: 'Security Warning',
+						detail:
+							[
+								`Error: ${error}`,
+								'',
+								`Subject: ${certificate.subjectName}`,
+								'',
+								`Issuer: ${certificate.issuerName ?? 'UNKNOWN'}`,
+								`- Organisations: ${certificate.issuer.organizations.join(', ')}`,
+								`- Organisation units: ${certificate.issuer.organizationUnits.join(', ')}`,
+								`- Country: ${certificate.issuer.country}`,
+								`- State: ${certificate.issuer.state}`,
+								`- Locality: ${certificate.issuer.locality}`,
+								'',
+								`Fingerprint: ${certificate.fingerprint}`,
+								'',
+								`Valid from: ${new Date(certificate.validStart * 1_000).toLocaleDateString()}`,
+								`Valid until: ${new Date(certificate.validExpiry * 1_000).toLocaleDateString()}`,
+								'',
+								'Do you trust this certificate?',
+							].join('\n'),
+						buttons: ['Yes', 'Cancel'],
+					})
+					openMessageBoxes.set(certificate.fingerprint, messageBox)
+				}
+				messageBox.then(({ response }) => {
+					const isFirst = openMessageBoxes.delete(certificate.fingerprint)
+					if (response === 0) {
+						if (isFirst) {
+							setAppConfig('trustedFingerprints', [certificate.fingerprint, ...trustedFingerprints])
+						}
+						callback(true)
+					} else {
+						callback(false)
+					}
+				})
+			}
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -218,49 +218,58 @@ app.whenReady().then(async () => {
...
+	const trustedFingerprints = getAppConfig('trustedFingerprints') ?? []
 	app.on('certificate-error', (event, webContents, url, error, certificate, callback) => {
 		event.preventDefault()
 
 		if (isLinux) {
-			return callback(process.env.NODE_ENV !== 'production')
+			if (trustedFingerprints.includes(certificate.fingerprint)) {
+				callback(true)
+			} else {
+				dialog.showMessageBox(mainWindow, {
```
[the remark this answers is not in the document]
ShGKme: Or in the Talk window

image

src/main.js
```diff
@@ -218,49 +218,58 @@ app.whenReady().then(async () => {
...
+							`Valid from: ${new Date(certificate.validStart * 1_000).toLocaleDateString()}`,
+							`Valid until: ${new Date(certificate.validExpiry * 1_000).toLocaleDateString()}`,
+							'',
+							'Do you trust this certificate?',
+						].join('\n'),
+					buttons: ['Yes', 'Cancel'],
+				}).then((res) => {
+					if (res.response === 0) {
+						trustedFingerprints.push(certificate.fingerprint)
+						setAppConfig('trustedFingerprints', trustedFingerprints)
```
ShGKme, line 256: At this moment there is a chance that the certificate was already added from another dialog.
Let's only add a new certificate if it is not added yet in parallel.

ShGKme: Maybe this is `setAppConfig` problem, but it only saves the config when the set value is actually new. And it compares it by reference. Setting the same array with a new pushed certificate has no effect.

As a result, it only works when `trustedFingerprints` was not defined before.

Having an empty list, or a list with old fingerprints prevents from adding a new one.

```json
{
  "trustedFingerprints": [],
  "//": "or",
  "trustedFingerprints": ["...old fingerprint..."]
}
```

ShGKme: A simple option here is creating a new array instead of a push.

Alternative - adjust `setAppConfig` to compare values by value, not reference. For example, via `JSON.stringify`.

src/main.js
```diff
@@ -218,49 +218,58 @@ app.whenReady().then(async () => {
 		}
 	})
 
+	const trustedFingerprints = getAppConfig('trustedFingerprints') ?? []
 	app.on('certificate-error', (event, webContents, url, error, certificate, callback) => {
 		event.preventDefault()
 
 		if (isLinux) {
-			return callback(process.env.NODE_ENV !== 'production')
+			if (trustedFingerprints.includes(certificate.fingerprint)) {
```
ShGKme, line 226: Let's always get the `trustedFingerprints` from the app config in the event handler.

Currently, there is no error from getting the config value only once, because `trustedFingerprints` variable is changed only here. But if there is another source of changing the config, here we'll have an outdated value. For example, we may add a fingerprint reset option in the future.

`getAppConfig` is cheap, it uses in-memory cache.

ShGKme (changes requested):
Thanks a lot for your contribution! We'll definitely get it in.

It works but misses some edge cases. Could you adjust it according to the review?

src/main.js
```diff
@@ -218,49 +218,68 @@ app.whenReady().then(async () => {
...
+					})
+					openMessageBoxes.set(certificate.fingerprint, messageBox)
+				}
+				messageBox.then(({ response }) => {
+					if (response === 0) {
+						if (openMessageBoxes.delete(certificate.fingerprint)) { // only updates the AppConfig once
+							setAppConfig('trustedFingerprints', [certificate.fingerprint, ...trustedFingerprints])
+						}
+						callback(true)
+					} else {
```
ShGKme, line 268: I haven't tested it yet, but it looks like every `certificate-error` creates a new promise resolve handler to add the certificate to the config. And, although the dialog is created only once, the app config is updated as many times as there was an error.

Promise handler should be added only once, when the dialog is created, and only the `callback` need to be called on every error.

Captainpast: No, `setAppConfig` is only called once, because its only called if [`Map.prototype.delete()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete#return_value) returns **true** and this is only the case if the element is still in the `Map`.
So after the first callback is deleting the _fingerprint_ from the `Map` it's also adding the _fingerprint_ to _AppConfig_ and the every other `delete()` return **false**.

ShGKme: Ah, yes

src/main.js
```diff
@@ -218,49 +218,68 @@ app.whenReady().then(async () => {
...
+						callback(false)
+					}
```
ShGKme, line 270: This works well when the certificate is accepted:
- Promise is deleted
- Config is updated
- Certificate is accepted

But when it isn't (User pressed "Cancel"), just nothing happens.
The app just doesn't work, all the following requests are blocked.
And there are no dialogs anymore, because it is still in `openMessageBoxes`.
The user must restart the app.

And maybe the user did it just by accident.

It should:
1. Either show the dialog again on the following requests
2. Or prevent the user from using the app

IMO, 1 is much simpler. 

Let's always remove the promise once it is resolved. Not only if it was resolved with a 0 response.

Issue #387, opened 2023-10-23
Application protocol


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**

Register application protocol to allow:
- Open the app
- Open a specific chat
- Join a call
- Decline a call

**Describe the solution you'd like**

Register an app protocol like `nextcloudtalk://chat/:id`.

**Describe alternatives you've considered**

-

**Additional context**

It would allow:
- Have a button "Open in Nextcloud Talk Desktop" on the Web UI
- Open Nextcloud Talk Desktop from the Nextcloud Client, for example, by the Icon, of from the call notification:
  ![image](https://github.com/nextcloud/talk-desktop/assets/25978914/d7842075-3dd9-4d96-8180-9cbe296ef3a9)

Originally mentioned in: https://github.com/nextcloud/talk-desktop/issues/386

neufeind:
Related issue in the talk-server (spreed) issues: https://github.com/nextcloud/spreed/issues/13902

Issue #9, opened 2023-02-06
Meta issue: Security improvement


ShGKme:
Checklist:

- [x] Remove `SameSite=Lax -> SameSite=None` cookies patching (fixed in: https://github.com/nextcloud/talk-desktop/pull/22)
- [x] ~~#18~~
- [ ] Follow [Electron / Best Practices / Security](https://www.electronjs.org/docs/latest/tutorial/security)
  - [x] 1. Only load secure content
  - [x] 2. Do not enable Node.js integration for remote content
  - [x] 3. Enable Context Isolation
  - [x] 4. Enable process sandboxing
  - [ ] 5. Handle session permission requests from remote content
  - [x] 6. Do not disable `webSecurity` (https://github.com/nextcloud/talk-desktop/pull/22)
  - [ ] 7. Define a Content Security Policy
  - [x] 8. Do not enable `allowRunningInsecureContent`
  - [x] 9. Do not enable experimental features
  - [x] 10. Do not use `enableBlinkFeatures`
  - [x] 11. Do not use `allowpopups` for `WebViews`
  - [x] 12. Verify `WebView` options before creation
  - [x] 13. Disable or limit navigation
  - [x] 14. Disable or limit creation of new windows
  - [ ] 15. Do not use `shell.openExternal` with untrusted content
  - [x] 16. Use a current version of Electron
  - [ ] 17. Validate the sender of all IPC messages
  - [x] 18. Avoid usage of the `file://` protocol and prefer usage of custom protocols (https://github.com/nextcloud/talk-desktop/pull/1204)
  - [ ] 19. Check which `fuses` you can change
  - [x] 20. Do not expose Electron APIs to untrusted web content

PR #1204, opened 2025-04-09
chore: add internal app protocol

ShGKme:
## ☑️ Resolves

* For https://github.com/nextcloud/talk-desktop/issues/9
  * 18: https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols
* Base for:
  * Some other improvements
  * https://github.com/nextcloud/talk-desktop/issues/387

Note, it's only an internal protocol. It is not exposed to the system (yet).

### Before

Resources where loaded via `file://` protocol and absolute file path.

### After

There is an internal application protocol with the following URLs:
- `nctalk://app/` - Application files
  - `nctalk://app/welcome_window/index.html` - welcome page
  - `nctalk://app/talk_window/index.html` - Talk
  - `nctalk://app/talk_window/index.js` - static file
  - etc.
- `nctalk://call/{token}/` - Link to a specific conversation
  - Redirects to `nctalk://app/talk_window/index.html#/call/{token}/`

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -20,17 +20,11 @@ export default [
 		languageOptions: {
 			globals: {
 				// Electron Forge build vars
-				AUTHENTICATION_WINDOW_WEBPACK_ENTRY: 'readonly',
 				AUTHENTICATION_WINDOW_PRELOAD_WEBPACK_ENTRY: 'readonly',
 				CALLBOX_WINDOW_PRELOAD_WEBPACK_ENTRY: 'readonly',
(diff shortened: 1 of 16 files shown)


Antreesy (approved):
Works fine, smoke tested with chats, calls, file downloads:
image

Couldn't find, if `src/patchers/@nextcloud/router.js` may fault somewhere

ShGKme:
> Works fine, smoke tested with chats, calls, file downloads

Thanks!

PR #1236, opened 2025-04-23
fix: avoid preflight (OPTIONS) request by using internal reverse proxy

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/1234
* The app is on `nctalk://app` host
* The server is on `https://some.nextcloud.ltd`
* Request from the app to the server is a cross-origin request (forbidden)
* In the past was "allowed" by changing the server response
  * This still makes an unnecessary server request
* Now it redirects to internal `ncapp://api/https://some.nextcloud.ltd` and proxies the request
  * It still has an ugly redirect
  * But no `OPTIONS` requests to the server

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/src/app/appProtocol.ts
+++ b/src/app/appProtocol.ts
@@ -16,6 +16,7 @@ protocol.registerSchemesAsPrivileged([
 			secure: true,
 			allowServiceWorkers: true,
 			supportFetchAPI: true,
+			corsEnabled: true,
 		},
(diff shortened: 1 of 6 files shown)


Antreesy (approved):
Tested

Issue #1233, opened 2025-04-22
Don't request backup status when user status setting dialog is not open


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

Requesting the backup status on every heartbeat allows opening the setting immediately knowing the actual status but creates an unnecessary load on the server.

PR #1237, opened 2025-04-23
perf(user-status): do not request backup status until user status form open

ShGKme:
### ☑️ Resolves

* Fix: https://github.com/nextcloud/talk-desktop/issues/1233
* Before backup status was requested on each heartbeat
* After it is requested only before opening the user status form
* Also:
  * Removed user status request from User Menu's avatar — use preload status
  * Fix the backup status design, as it is now more visible (I'm fine with dropping this commit)
  * Refactored a bit:
    * Decompose predefined statuses from the user status store
    * Use `useBrowserStorage` for LS caching

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image


Antreesy (approved):
Tested


B | A
-- | --
2025-04-23_18h48_23 | 2025-04-23_19h21_34

Issue #1209, opened 2025-04-11
Cannot download flatpak client


gewinh:
The link https://github.com/nextcloud-releases/talk-desktop/releases/latest/download/Nextcloud.Talk-linux-x64.flatpak to down the linux flatpak talk client  results in a 404 error ie the page does not exist.  Tried both chrome and brave got the same result when clicking on the link under install, linux on the home page in github.

nickvergessen:
@ShGKme some how the . was removed

Javohle:
Seems all download links leads to nowhere 404

stefan6419846:
The downloads are available in theory, but the filename changed from `Nextcloud.Talk` to `NextcloudTalk`, thus breaking the links in the README.

You can find the latest release builds at https://github.com/nextcloud-releases/talk-desktop/releases/tag/v1.1.6

Javohle:
Thank youfor information)

ShGKme:
Renamed

ShGKme:
> [@ShGKme](https://github.com/ShGKme) some how the . was removed

I missed that we used 2 different sanitized names in different places

Javohle:
Such a question, google tells a little, talk desktop client on win 7 installs from msi but dont start, fails to installs from exe, can i found some information what libraries or something is needed for it to work? in squirrell logs nothing too so i dont know is that an only that example of win7 problem or it is more global? old answers in google tells people just run talk.exe as portable

Javohle:
> [@Javohle](https://github.com/Javohle) The minimal system requirement is Windows 10.
> 
> Extended support of Microsoft Windows 7 ended on 14 Jan 2020 and Paid Extended Security Updates on Jan 2023. It doesn't receive security updates anymore, and so do other used tools supporting Windows 7, for example, Electron or Node.js.

Sad, but true

gewinh:
Again just for info for ubuntu uses, when the flatpak update happens to go to 21.0.2, a reboot was required to get talk to run again.

PR #1250, opened 2025-04-29
build: use dot in sanitized dist name

ShGKme:
### ☑️ Resolves

* Fix https://github.com/nextcloud/talk-desktop/issues/1209
  * It was fixed because I renamed files manually after the build 🙈
* Regression from https://github.com/nextcloud/talk-desktop/pull/1193
* There were different sanitized name formats...
  * In dist files we used `.` as a replacement of non-alphanumeric characters
  * In other places - non-alphanumeric characters are just removed


Antreesy (approved):
Looks valid

DorraJaouad (approved):

PR #1261, opened 2025-05-05
fix(app): rework networking to serve app over `https://<server_url>` origin instead of `nctalk://` or `file://`

ShGKme:
## Description

Rework networking ~~one more~~ the last time.

Fixes:
- Invalid `User-Agent`
- Broken caching
- Double requests in devtools
- vue-devtools support

Improves and simplifies networking in general.

### Before

Previous implementations:

1. Serving the app from `file://` protocol
   - ✅ Simple default solution
   - ✅ No double requests in devtools (only in development)
   - ✅ Vue-Devtools work
   - ⚠️ Requires manual Cookies management
   - ⚠️ Requires `@nextcloud/router` patching and some hacks to change the origin
   - ⚠️ ON DEV: Requires responses patching for CORS
   - ❌ Not recommended by Electron security guidelines
2. Serving from `nctalk://` + patching OPTIONS for CORS
   - ✅ Follows Electron security guidelines
   - ⚠️ Still requires manual Cookies management
   - ⚠️ Still requires `@nextcloud/router` patching and some hacks to change the origin
   - ⚠️ Every request is shown twice in devtools (OPTIONS + request)
   - ❌ Vue-Devtools do not work
   - ❌ Performance issue — it makes CORS requests on production
   - ❌ Requires patching responses in production, may lead (it did) to broken requests when some custom headers are forgotten
3. Serving from `nctalk://` + redirect to internal reverse proxy to Nextcloud instance
   - ✅ Follows Electron security guidelines
   - ✅ No manual Cookies management
   - ⚠️ Requires `@nextcloud/router` patching and some hacks to change the origin
   - ⚠️ Every request is shown twice in devtools (request + redirect)
   - ⚠️ Every request is 0b size in devtools
   - ❌ Vue-Devtools do not work
   - ❌ Somehow cache is not working anymore
   - Note: `axios` cannot make a request to a custom protocol directly

### After

New implementation: serving from `https://<server_url>` with internal proxy to the Nextcloud instance

- ✅ Follows Electron security guidelines
- ✅ Almost equivalent to the Web Client
- ✅ Vue-Devtools work
- ✅ Does not require `@nextcloud/router` patching and hacks
  - Some internal image/sound resources loading still uses, but it will be done in a follow-up
  - Additional `Authentication` header can now be omitted in requests with Cookies (will be done in follow-up)
- ✅ No drawbacks from the methods above
- ⚠️ Every request is still 0b size in devtools :(
  - Electron bug?
- ⚠️ Connecting to servers over `http://` is currently disallowed
  - Doable, but makes little sense

## How it works

- The app is served from:
  - `https://app/*` - windows without authentication
  - `https://<server_origin>/talk_desktop__window_<name>/*` - windows with authentication
  - `/talk_desktop__dist/*` - assets and chunks
- Other requests, including `https://<server_url>/*`, are proxied as it is, except:
  - Overriding `User-Agent` header from Electron's default to Nextcloud-Talk
  - Adding `Authentication` header for OCS API requests to the current Nextcloud instance 

Note: `<server_origin>` is only the origin without a base path.

Because Login and Talk windows are now served from different hosts, app data cannot be shared, and now is shared via the main process.

It's quite dirty at the moment — to be refactored in a follow-up...

## Follow-ups

1. Get rid of `@nextcloud/router` patching completely - handle resource loading in the app protocol handler
2. Improve `AppData` management
3. Remove `Authentication` header from requests where session in a cookie can be used
4. Check the bug with 0b size of the requests in devtools

## Test cases

- [x] Login to an account
- [x] Opening Talk window
- [x] Opening Talk Window after restart
- [x] Logout
- [x] Development with reload on change
- [x] Production build
- [x] Help window, upgrade window
- [x] Copy link actions
- [x] Sending files
- [x] Receiving files and previews
- [x] File picker dialog
- [x] Calendar
- [x] Sounds:
  - [x] On a new message
  - [x] On a callbox
  - [x] On waiting
- [x] Notifications with notify_push
- [x] General call without HPB
- [x] General call with HPB
- [x] Screensharing
- [x] Virtual background
- [ ] E2EE calls (if Virtual background works - it should as well)
- [x] Federated
- [ ] On base path like `/nextcloud/`
- [ ] Cross-platform
  - [x] Windows
  - [x] macOS
  - [ ] Linux
- ❌ Over http:// - not supported at the moment


src/app/utils.ts
```diff
@@ -84,13 +85,42 @@ export function buildTitle(title?: string) {
...
+ */
+export function isInternalUrl(url: string | URL) {
+	if (typeof url === 'string') {
+		url = new URL(url)
+	}
+
+	const internalOrigin = appData.serverUrl ? new URL(appData.serverUrl).origin : APP_ORIGIN
+
+	return url.origin === internalOrigin && url.pathname.startsWith('/talk_desktop__')
+}
```
ShGKme, line 117: Because the app is host now on `https://<server_origin>`, internal URLs can only be distinguished by the path.

This is why all the assets where moved to `/talk_desktop__*` sub directories.

src/app/utils.ts
```diff
@@ -96,5 +97,30 @@ export function getWindowUrl(window: typeof windows[]) {
...
+ * Check weather a link is an internal application link
+ *
+ * @param url - URL
+ */
+export function isInternalUrl(url: string | URL) {
+	if (typeof url === 'string') {
+		url = new URL(url)
+	}
+
+	const internalOrigin = appData.serverUrl ? new URL(appData.serverUrl).origin : APP_ORIGIN
```
Antreesy, line 114: can it be moved out of function calls (here and at L100) and reused?

Antreesy (approved):
Tested in call and with some basic functionality
Seems to work🤞🏽

src/app/utils.ts
ShGKme: In a simple way - no. The value can be changed in runtime, e.g., you logged out and logged in again. So it at least must be in the app data.

Will refactor it much when we add Multi-Account Support

PR #1284, opened 2025-05-12
chore: update pull_request_template.md

ShGKme:
### ☑️ Resolves

- `* Issue #…` doesn't set PR as a fix to an issue, only creates a link
- `Tasks` are not used in most of the cases:
  - Not relevant for fixes and chore
  - When needed for large features with many steps, can be described in a corresponding issue

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,13 +1,9 @@
 ### ☑️ Resolves
 
-* Issue #…
+- Fix: #…
 
 ### 🖼️ Screenshots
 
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Fine with that approach. Not sure if it should be just `Fix #...`

ShGKme:
> Not sure if it should be just `Fix #...`

What do you mean?

ShGKme:
Changing `issue` to `fix` required to connect a PR and an issue

PR #1299, opened 2025-05-16
fix(title-bar): remove logo click in favor of new home button

ShGKme:
### ☑️ Resolves

- Not needed with the new dashboard and explicit Home button


Antreesy (approved):

nickvergessen:
We shouldn't remove a user muscle memory unless there is a reason to break it and i don't see it here?
And e.g. the logo is visible even when the left sidebar is folded?

ShGKme:
I'm unsure if people even knew about this feature. Which even didn't exist on macOS.

PR #1391, opened 2025-07-21
fix: missing l10n due to early @nextcloud/l10n import

ShGKme:
### ☑️ Resolves

- Fix: some UI elements are not translated
- Regression from https://github.com/nextcloud/talk-desktop/pull/1375

`@nextcloud/l10n` should never be imported before the web page is set up, otherwise language might not be applied during init.

image

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


Antreesy (approved):
Lgtm

Issue #1402, opened 2025-07-30
Presence status incorrectly changes to offline when Talk is not focused but PC is active


Taio2003:
Steps to reproduce
Open Nextcloud Talk (v1.2.5) on Windows 11 or macOS.

Keep the Talk window open but in the background or minimized.

Actively use the PC (typing, moving the mouse, etc.) without interacting directly with the Talk window.

Observe that the user status randomly switches to offline or "unavailable" prematurely.

Expected behaviour
The user should remain online as long as they are actively using the PC, even if the Talk window is not in the foreground. The status should only switch to "unavailable" or offline after a defined period of actual user inactivity (no mouse or keyboard input).

Actual behaviour
The user is marked as offline or "unavailable" even when actively using the PC, as long as Talk is open but not focused. The status changes happen randomly and too early.

Diagnosis and logs
<details> <summary>Diagnosis report</summary>
The issue occurs on both Windows 11 and macOS with Nextcloud Talk version 1.2.5. The diagnostic report from the About window can be provided if needed.

</details> <details> <summary>Client logs</summary>
No obvious errors appear in the developer console (Ctrl+Shift+I / Cmd+Option+C → Console) during the status change events.

pgsql
Copia
Modifica
[INFO] Connection stable
[INFO] User presence updated: offline (unexpectedly)
No crashes or network errors related to presence or connection loss have been detected.

</details>

ShGKme:
We can reproduce the issue.

PR #1405, opened 2025-07-31
fix(user-status): heartbeat is not sent when (in)activity doesn't change

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud/talk-desktop/issues/1402
- Instead of recursively schedule heartbeating, only a single heartbeat sending was scheduled
- Rename a function to clarify it


nickvergessen (approved):

Antreesy:
Tested 🦥

ShGKme:
Thanks!

PR #1424, opened 2025-08-22
chore: update dependencies

ShGKme:
- Mostly minor bump
- `npm audit fix`
- Bump Electron from 36 to 37
  - No relevant breaking changes: https://releases.electronjs.org/release/v37.0.0

### DIFF
--- a/src/talk/renderer/Viewer/ViewerHandlerBase.vue
+++ b/src/talk/renderer/Viewer/ViewerHandlerBase.vue
@@ -38,7 +38,7 @@ const readyToShow = computed(() => !props.loading && !props.error)
 	<div class="viewer-wrapper">
 		<NcEmptyContent
 			v-if="loading"
-			:name="t('talk_desktop', 'Loading …')"
+			:name="t('talk_desktop', 'Loading …')"
 			class="viewer-wrapper__empty-content delayed-appear"
 			data-theme-dark>
 			<template #icon>
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@nextcloud/browser-storage": "^0.4.0",
         "@nextcloud/capabilities": "^1.2.0",
         "@nextcloud/event-bus": "^3.3.2",
-        "@nextcloud/files": "^3.10.2",
+        "@nextcloud/files": "^3.12.0",
         "@nextcloud/initial-state": "^2.2.0",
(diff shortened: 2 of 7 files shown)


src/talk/renderer/Viewer/ViewerHandlerBase.vue
```diff
@@ -38,7 +38,7 @@ const readyToShow = computed(() => !props.loading && !props.error)
 	<div class="viewer-wrapper">
 		<NcEmptyContent
 			v-if="loading"
-			:name="t('talk_desktop', 'Loading …')"
+			:name="t('talk_desktop', 'Loading …')"
```
Antreesy, line 41: So, we're touching the strings?

Antreesy (approved):
LGTM

src/talk/renderer/Viewer/ViewerHandlerBase.vue
ShGKme: @susnux Do you know if this auto fix also keeps translations in Transifex for apps, or it only works in libs? 

If it doesn't, we probably should remove auto-fixing in this rule by default, as it may break a lot of translations...

susnux: No it will only use the fuzzy translation finder meaning it will be marked as not reviewed and needs a new review.
But this is expected.

Also those fixes are expected - using a plain space here is the same as any other typo as per our localization guide. So as long as its not string freeze it should be ok to do so and fix this.

ShGKme: The only rule I know is this one: https://github.com/nextcloud/server/wiki/Translation-knowledge-(valid-for-whole-nextcloud-project)/d45f1d248fdc6b1509337ebb0017653649a8f440#ellipsis
- It asks to use space in a sentence and no space when shortening a word
  - Example: 'Uploading …', same as Loading - is a full word.
- It doesn't ask to use non-breaking space...

ShGKme:
Removed `eslint-config` update. Kept only simple dependencies update without changes in the code.

src/talk/renderer/Viewer/ViewerHandlerBase.vue
susnux: But typographical there must be a non breaking space as otherwise the word wrap would be wrong.
Similar like between numbers and units etc

Issue #1305, opened 2025-05-20
MSI installer should not install the program in a subpath containing the app version


cm-schl:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
Using the MSI package to install Nextcloud Talk Desktop in an Windows environment that requires admin rights to install software I saw that on every update the new version is installed in a path that also contains the version of the app.
Example: `C:\Program Files\Nextcloud Talk\app-1.1.9`
Because the MSI package also registers a new entry inside the list of installed software inside the system settings of Windows this leads to multiple entries.

**Describe the solution you'd like**
I would like to propose that the app gets installed without app version in the path and overwrites the previous version of the program. The same should happen in the system settings: only one entry should show up when Nextcloud Talk Desktop is installed on the machine.

ShGKme:
The app is installed into the `C:\Program Files\Nextcloud Talk`, not into a subfolder. There are subfolders with specific versions, but the executable used in all the shortcuts is still `Nextcloud Talk.exe` in the root. This is normal and expected behavior.

![Image](https://github.com/user-attachments/assets/ae85c6e5-ab9e-40c5-b8c8-37cad364957c)

However, having multiple entries in Windows is not expected.

![Image](https://github.com/user-attachments/assets/06d0284a-1363-4ae5-8589-cc6a24ced3ed)

This happens not due to the upgrade mechanism with subfolders, but due to undefined update GUID.

PR #1435, opened 2025-08-29
fix(build): msi update installs as a new app

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud/talk-desktop/issues/1305
- Set a specific Upgrade_Code => installing a new version 
- Generate UUIDv6 so it stays the same for the same app id
  - Branded clients have new app id => new upgrade code


Antreesy (approved):
Looks good

PR #1442, opened 2025-09-02
fix: defer Vue app import until webpage setup is done

Antreesy:
### ☑️ Resolves

- Fix interface errors caused by early import

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

- [x] Checked stable1, no backport needed


ShGKme (approved):
Thanks!

ShGKme:
> * Checked stable1, no backport needed

I'd backport anyway, seems that it may break later.

PR #1451, opened 2025-09-04
build: add server@32 styles

ShGKme:
### ☑️ Resolves

- Add support for server32 styles (with REUSE.toml instead of reuse/.dep5)
- Remove stable31 styles (not compatible with Talk Desktop v2)
- Fixed some broken styles

### New styles

image

image

### Additionaly fixed

image


src/callbox/renderer/CallboxApp.vue
```diff
@@ -114,6 +114,23 @@ function dismiss() {
 	</AppWindow>
 </template>
 
+<style>
+:root {
+	--join-call-background-color: var(--color-border-success);
```
Antreesy, line 119: should be used `var(--color-element-success);` instead of border, also in spreed repo

Antreesy (approved):
Otherwise good

src/callbox/renderer/CallboxApp.vue
ShGKme: I'll fix when it is fixed in `spreed` to be consistent. Or when it is fixed globally.

PR #1459, opened 2025-09-09
add detailed install instructions for linux flatpak

alexanderdd:
tested - works


nickvergessen (approved):

nickvergessen:
> I don't have the right to merge

You have, if CI is green. It complained because your commit was not following our pattern, so I added the key word `docs:` to the beginning and tada it's green and now you should be able to press the button

Issue #1476, opened 2025-09-20
Only the first 1-2 seconds of the video are played.


ArthurSib:
## Steps to reproduce
1. Open any chat with video.
2. Click play video.

### Expected behaviour
Playing videos entirely with the ability to position on a timeline. 

### Actual behaviour

Playing the video for 1-2 seconds and stop it and without the possibility of positioning on a timeline.

<img width="1393" height="845" alt="Image" src="https://github.com/user-attachments/assets/52fd9db2-9bc4-4165-b82e-de2c8e060018" />

## Diagnosis and logs

<details>
<summary>Diagnosis report</summary>

### Diagnosis report

| **Nextcloud Talk Desktop**      |                                                                                            |
| ------------------------------: | ------------------------------------------------------------------------------------------ |
| **Version**                     | v1.2.6                                                                                     |
| **Built-in Talk version**       | v21.1.4                                                                                    |
| **Release channel**             | stable                                                                                     |
| **Operating system**            | Linux 6.14.0-29-generic (#29~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Aug 14 16:52:50 UTC 2) |
| **Executable Path**             | /app/lib/com.nextcloud.talk/Nextcloud Talk                                                 |
| **Using Wayland**               | ❌ no                                                                                       |
| ****Connected to****            | -                                                                                          |
| **Nextcloud version**           | 31.0.9                                                                                     |
| **Nextcloud Talk version**      | 21.1.5                                                                                     |
| **`notifications` app enabled** | ✅ yes                                                                                      |
| **`notify_push` app enabled**   | ✅ yes                                                                                      |

#### Application config

```json
{
  "launchAtStartup": false,
  "theme": "dark",
  "systemTitleBar": true,
  "monochromeTrayIcon": false,
  "zoomFactor": 1,
  "playSoundChat": "respect-dnd",
  "playSoundCall": "respect-dnd",
  "enableCallbox": "respect-dnd",
  "secondarySpeaker": false,
  "secondarySpeakerDevice": null,
  "trustedFingerprints": []
}
```

It can be found in the About window (open from the menu or via <kbd>F1</kbd> and click on "Copy report" button).

</details>

**Additional context**

Firefox and Chromium do not have this issue.

ArthurSib:
@ShGKme, in version v2.0.2 the problem still reproducible.
[...]

Issue #1461, opened 2025-09-11
Suddenly getting Network Error after opening video file


NikoKS:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open Nextcloud Talk Desktop
2. Open a video attachment
3. Try sending message

### Expected behaviour
message sent

### Actual behaviour
keep getting this error
```json
{
    "message": "Network Error",
    "name": "AxiosError",
    "stack": "AxiosError: Network Error\n    at C.onerror (https://cloud.example.com/talk_desktop__window_talk/index.js:2:9523664)\n    at Ci.request (https://cloud.example.com/talk_desktop__window_talk/index.js:2:9531399)\n    at async l (https://cloud.example.com/talk_desktop__dist/chunks/3079.js?v=98d86632ebfae54d64fc:1:1113103)",
    "config": {
        "transitional": {
            "silentJSONParsing": true,
            "forcedJSONParsing": true,
            "clarifyTimeoutError": false
        },
        "adapter": [
            "xhr",
            "http",
            "fetch"
        ],
        "transformRequest": [
            null
        ],
        "transformResponse": [
            null
        ],
        "timeout": 0,
        "xsrfCookieName": "XSRF-TOKEN",
        "xsrfHeaderName": "X-XSRF-TOKEN",
        "maxContentLength": -1,
        "maxBodyLength": -1,
        "env": {},
        "headers": {
            "Accept": "application/json, text/plain, */*",
            "Content-Type": "application/json",
            "requesttoken": "",
            "X-Requested-With": "XMLHttpRequest"
        },
        "method": "put",
        "url": "https://cloud.example.com/ocs/v2.php/apps/spreed/api/v4/room/oijaoeyf/participants/state",
        "data": "{\"state\":0}",
        "allowAbsoluteUrls": true,
        "axios-retry": {
            "retries": 0,
            "shouldResetTimeout": false,
            "validateResponse": null,
            "retryCount": 0,
            "lastRequestTime": 1757580220971
        }
    },
    "code": "ERR_NETWORK"
}
```

## Diagnosis and logs
also got this log. not sure if related: 

- Not connected to signaling server yet, defer joining room
[...]

Issue #1429, opened 2025-08-26
Video Streaming Network error: Video playback stops after some time


Giorgio-Germani:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Try to send this video on a chat: https://cloud.reventor.eu/s/WEyBfa2sokqAF23
2. Video Stops playback afer a few seconds.

### Expected behaviour
Video should play completely

### Actual behaviour
Video stops playback after a few seconds

### Diagnosis report

| **Nextcloud Talk Desktop**      |                                                                         |
| ------------------------------: | ----------------------------------------------------------------------- |
| **Version**                     | v1.2.5                                                                  |
| **Built-in Talk version**       | v21.1.2                                                                 |
| **Release channel**             | stable                                                                  |
| **Operating system**            | Windows_NT 10.0.26100 (Windows 11 Pro)                                  |
| **Executable Path**             | C:\Users\Admin\AppData\Local\NextcloudTalk\app-1.2.5\Nextcloud Talk.exe |
| ****Connected to****            | -                                                                       |
| **Nextcloud version**           | 31.0.8                                                                  |
| **Nextcloud Talk version**      | 21.1.3                                                                  |
| **`notifications` app enabled** | ✅ yes                                                                   |
| **`notify_push` app enabled**   | ✅ yes                                                                   |

#### Application config

```json
{
  "launchAtStartup": false,
  "theme": "default",
  "systemTitleBar": false,
  "monochromeTrayIcon": false,
  "zoomFactor": 1,
  "playSoundChat": "respect-dnd",
  "playSoundCall": "respect-dnd",
  "enableCallbox": "respect-dnd",
  "secondarySpeaker": false,
  "secondarySpeakerDevice": null,
  "trustedFingerprints": []
}
```

Fuseteam:
@ShGKme ubuntu 24.04

ShGKme:
@Fuseteam Do you have the same problem?
1. You start playing the video
2. It stops at a random/specific moment

Does it happen all the time, or occasionally, for example, after the client was open for a long time (several days)?

NikoKS:
[...]

Issue #1401, opened 2025-07-29
Video file download/streaming is incomplete after first file, only works after restarting app (Windows Desktop)


auslinx73:
## Steps to reproduce

1. Start Nextcloud Talk Desktop (v1.2.5, Windows 11).
    
2. Upload a video file to a chat (Talk conversation).
    
3. Play the video — it loads and plays completely.
    
4. Upload another video file (or the same file again).
    
5. Try to play it — only part of the file (10–30%) loads, then stops and never finishes.
    
6. If you quit and restart the app, the first file will again load 100%. Any next video will get stuck after ~10–30%.
    

### Expected behaviour

All video files should fully load and play, regardless of how many videos are played or uploaded, just like in the web Talk interface.

### Actual behaviour

- The **first video file** played after app start loads and plays fine (100%).
    
- Any **subsequent video** only loads up to 10–30% and never finishes loading. Playback stops.
    
- This is **fixed only after restarting the app**.
    
- **In the web interface (browser) and on Android** everything works as expected, without this bug.
    

## Diagnosis and logs

<details> <summary>Diagnosis report</summary>

pgsql

КопироватьРедактировать

| **Nextcloud Talk Desktop**      |                                                                         | | ------------------------------: | ----------------------------------------------------------------------- | | **Version**                     | v1.2.5                                                                  | | **Built-in Talk version**       | v21.1.2                                                                 | | **Release channel**             | stable                                                                  | | **Operating system**            | Windows_NT 10.0.26100 (Windows 11 Home Single Language)                 | | **Executable Path**             | C:\Users\yotam\AppData\Local\NextcloudTalk\app-1.2.5\Nextcloud Talk.exe | | **Nextcloud version**           | 31.0.7                                                                  | | **Nextcloud Talk version**      | 21.1.2                                                                  | | **`notifications` app enabled** | ✅ yes                                                                   | | **`notify_push` app enabled**   | ❌ no                                                                    |

</details> <details> <summary>Client logs</summary>

Press:

- **🪟 Windows:** <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd>
    
- **🍎 macOS:** <kbd>Cmd</kbd>+<kbd>Option</kbd>+<kbd>C</kbd>
    
- **🐧Linux:** <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd>
    

Or use the context menu to open the developer tools, then go to the Console tab.

[...]

PR #1478, opened 2025-09-22
fix(app): support network streams

ShGKme:
### ☑️ Resolves

- Might fix:
  - https://github.com/nextcloud/talk-desktop/issues/1476
  - https://github.com/nextcloud/talk-desktop/issues/1429
  - https://github.com/nextcloud/talk-desktop/issues/1461
  - https://github.com/nextcloud/talk-desktop/issues/1401


ShGKme:
According to tests with @miaulalala it doesn't fix the issue :(

But from what I understand, we still should have this option and it doesn't break anything.

Let's try to see if it helps at least for some users.

Issue #1484, opened 2025-09-30
Windows Firewall Rule


outnsp:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
Windows firewall rule missmatch

**Describe the solution you'd like**
Remove the version number from the file path. When the Windows firewall rule is created by confirming the pop‑up, the exact file path is recorded in the rule. After a Nextcloud Talk update, the file path no longer matches the path stored in the firewall rule.

ShGKme:
I guess this is about .msi installer, correct?

> Remove the version number from the file path

Doesn't it work with the following path?
`C:\Program Files\Nextcloud Talk\Nextcloud Talk.exe`

Nimbusaro:
Hi @ShGKme ,

I just checked with our Windows PC fleet, where Talk is installed via the MSI package. When a user opens a video call for the first time, they get a Windows Firewall prompt. Since they’re not admins, they can only cancel. This automatically creates a 'reject' rule in the firewall, and we can see that it includes the path with Talk version number.

<img width="1330" height="128" alt="Image" src="https://github.com/user-attachments/assets/237dc192-a6f2-4bda-91ce-f3533376180e" />

I deleted these rules and tried manually creating an 'allow' rule for the path 'C:\Program Files\Nextcloud Talk\Nextcloud Talk.exe' (path verified and correct), but it doesn’t work—Windows Firewall still triggers the prompt and recreates the 'reject' rule.

<img width="419" height="365" alt="Image" src="https://github.com/user-attachments/assets/b00f1d43-f85b-474f-b4dd-b5c193541d4b" />

(in english: "Do you want to allow network to access this app - Windows firewall has blocked Nextcloud Talk features")

<img width="1543" height="110" alt="Image" src="https://github.com/user-attachments/assets/29044550-c55a-4fe4-9027-e6930cf6d823" />

ShGKme:
Thanks for the details, now it is clear

PR #1518, opened 2025-10-31
build(windows/msi): add Firewall exception during install

ShGKme:
## ☑️ Resolves

- Fix: https://github.com/nextcloud/talk-desktop/issues/1484

### 🏚️ Before - the first call asks for adding a rule

image

### 🏡 After - no prompt + the rule is added

image

### The path — to the latest installed version only (even on upgrade)

image


nickvergessen (approved):
Can't test, but screenshots look good

PR #1155, opened 2025-03-05
[WIP] chore(config): re-design 

ShGKme:
### ☑️ Resolves

* Issue #…

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

PR #1530, opened 2025-11-07
fix(talk/settings): adjust to new design

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud/server/issues/55667
- Ref: https://github.com/nextcloud/spreed/issues/16113
- Bye 🥲 https://github.com/nextcloud/talk-desktop/pull/1155
- Problems:
  - Having +1 structuring level is lost
  - ~~`NcSelect` for Nextcloud 30 design is not implemented yet~~
    - https://github.com/nextcloud-libraries/nextcloud-vue/pull/7838
  - ~~Zoom is draft - I need to think about it~~
  - ~~Radio groups don't look well with long text even on a normal screen~~
    - Replaced with select
  - I loved having icons :(

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


nickvergessen:
> Radio groups don't look well with long text even on a normal screen

Yeah, that immediately jumped my eye as well. Should stay with a dropdown there I think

ShGKme:
Full settings for the complete view:

image

kra-mo (changes requested):
Looks quite nice. Some smaller points of feedback:
- The reset split button could have a bit more left/right padding, enough to align the icon with trailing icons in `FormBox*` components.
- The two options related to secondary speakers should probably have their own group to indicate that they are more related to each other than any other two notification options. Even better would be if they could be in the same box, but not sure if that works with the split button.
- "Notifications and sounds" could be renamed to "Notifications & Sounds" for consistency
- "Theme" could be moved below "Appearance" for structure. 
- The zoom controls are indeed a bit tricky. For now, I'm thinking:

Image

This gets rid of the slider but it makes it so you don't need to move your mouse too much to click on plus after you clicked on minus and vice versa.

It would get rid of the keyboard shortcuts, but that's probably fine? Ideally, they would be in the actual Keyboard shortcuts section anyway, but I'm guessing we don't want to insert them there?

ShGKme:
> The reset split button could have a bit more left/right padding, enough to align the icon with trailing icons in FormBox* components.

🥲👍 

(Needs more hacking or new components or making our buttons FormBox compatible by default/additionally)

> The two options related to secondary speakers should probably have their own group to indicate that they are more related to each other than any other two notification options. Even better would be if they could be in the same box, but not sure if that works with the split button.

👍

> "Notifications and sounds" could be renamed to "Notifications & Sounds" for consistency

👍

> It would get rid of the keyboard shortcuts, but that's probably fine? Ideally, they would be in the actual Keyboard shortcuts section anyway, but I'm guessing we don't want to insert them there?

Possible with some additional work, will do it later. Though, makes more sense if we have many desktop-specific shortcuts.

> * "Theme" could be moved below "Appearance" for structure.

It cannot, because from the new settings design structure, a radio group has the same structuring level as a general group (it is also a group).

Level 1: Settings Section (labelled)
Level 2: A group (form group or radio group, labelled)
Level 3: Visual box - subgroup (buttons in the radio group, list of switches)

> The zoom controls are indeed a bit tricky. For now, I'm thinking:

Then:
- Manual input is lost - not possible to select a value in between anymore
- The button named as "Zoom" visually and from a11y has a "Reset" icon and acts like a reset, which is a semantic and accessibility issue
- Slider is quite a classic element for the zoom, does it look bad :( ?

ShGKme:
Using "Reload" icon for the secondary speaker reload, and restore icon instead if "reset" text for the zoom.

image

ShGKme:
- Rebased to resolved conflict after dependabot merges

kra-mo:
> Making it "below' doesn't actually make it structurally inside "Appearance". If we want to make it inside - we need 1 more structural level

Not semantically, but generally, it does. This type of organization is still an important part of the part of layouts like this imo and I used similar thinking to organize other settings dialogs.

> "Theme" is the setting most of the users will use - I'd prefer to keep it on top

I did also think about theme being a more common option, but I thought that it has such a distinctive selector and it would still be pretty high up (you never have to scroll to see it) that it wouldn't be an issue.

ShGKme:
Another proposal, keeping the reset button and with an button to edit the value manually.

image

kra-mo:
> We don't have a select in this design. The label will be outside.

That's fine for now as long as the label is aligned (as I've been nagging you about in DMs, sorry!)

> Either it will be wide (with an empty space after short "100%" string), or we have a short select with empty space in the row.

It being wide is OK. If there's gonna be whitespace either way, it might as well increase the click target instead of contribute to the empty void :)

> Also, the values are not beautiful to the end user

Oh no, as I said earlier, if we go with a dropdown, I would show only larger fixed, amounts in there. Browsers and other apps do this with increments. So for example:

50%
75%
100%
125%
150%
200%
300%

Then the buttons could still adjust the zoom level by 5% increments for finer adjustments.

> Another proposal, keeping the reset button and with an button to edit the value manually.

You know, that's fine by me, I like it. I'm still not entirely sure about the reset button as you can just select 100%, but it looks fine… And if there is no reset button, I'd also move the edit button to the right to keep the text aligned. But your call, really.

The only question: Would the edit button edit the value in-line? Would it spawn a dialog or just an input field nearby?

ShGKme:
> Oh no, as I said earlier, if we go with a dropdown, I would show only larger fixed, amounts in there. Browsers and other apps do this with increments. So for example:
> 
> 50% 75% 100% 125% 150% 200% 300%
> 
> Then the buttons could still adjust the zoom level by 5% increments for finer adjustments.

It had a different change for a reason.

Moving from 50% to 100% (+50%) makes everything twice as large, which is a significant change.
Moving from 400% to 450% (also +50%) is not very noticeable and feels like moving from 100% to 125%.

In the web-browsers it is always annoying that you can Zoom with a mouse/keyboard, and the zooming in is slower and slower while zooming out is faster and faster.

That is why we use the factor.

PR #1538, opened 2025-11-12
fix(globals): fallback to the file icon for unknown mime types

ShGKme:
### ☑️ Resolves

- Fix: handling an edge case when the mime type is unknown, fallback to files


Antreesy (approved):
Is remaining `return undefined` aligned with server implementation? Feels like we could also pass a file icon there

Antreesy (approved):

Issue #1540, opened 2025-11-13
Join a conversation from the same server by entering a direct link


Cris70:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
I am using Nextcloud Talk on linux, through the official flatpak package.
When a colleague sends me a Talk link, I cannot open it in the Nextcloud Talk desktop application, because it opens in the default browser. 
If I put the link in the search bar of the desktop application, it tries to create a new conversation with that link as name.
If I click on "Join open conversations", the conversation does not appear, even when it is already started (probably because I was not explicitly added as a participant).
I tried to create a custom link handler, but there are no documented command-line options. Trying to run the application from the CLI with the URL as parameter does not open the conversation.

**Describe the solution you'd like**
In order of preference, starting from the most preferred:
- A way to open a conversation directly in-app through the shared link (as Google Meet)
- A way to make the desktop app open when the URL is opened in a browser
- A way to pass the URL as parameter to the command line, so that I can create a custom handler

**Describe alternatives you've considered**
Alternatives (or, rather, work-arounds) I considered are already described above.

**Additional context**
Nextcloud Talk v2.0.2 on linux openSUSE Tumbleweed. I know v2.0.3 is available, but when I tried it it couldn't connect to the server (which is v31.0.10).

Cris70:
> > * A way to open a conversation directly in-app through the shared link (as Google Meet)
> 
> If you mean having a link/button that directly opens Talk Desktop, it is covered by:
> 
>     * [Application protocol for "Open in Talk Desktop" links #387](https://github.com/nextcloud/talk-desktop/issues/387)

No, what I mean is what you're proposing in your next message.

Regarding the implementation of the application protocol, that would be awesome, but I see the issue is stalled since 2023...

neufeind:
Partially related: https://github.com/nextcloud/spreed/issues/13902
That ticket is about opening certain URLs (maybe nextcloudtalk://... or so) with the app. When using a https-URL and sharing it with the app (on iOS/Android) or manually entering that URL (like proposed here) it would allow to join that discussion.

[...]

PR #1541, opened 2025-11-13
[DRAFT] feat(talk): add "Join by link" action

ShGKme:
### ☑️ Resolves

- Fix: https://github.com/nextcloud/talk-desktop/issues/1540
- Wording is WIP
- **Alternative: do it on the web, for example, on the Dashboard.**

### 🖼️ Screenshots

image

image

image

image

image

image

image

image

image

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -26,3 +26,16 @@ export function once<F extends (() => any) | ((...args: any[]) => void)>(func: F
 }
 
 /* eslint-enable @typescript-eslint/no-explicit-any */
+
+/**
+ * Create URL instance or return null if invalid
+ *
+ * @param url - URL string
+ */
+export function createURL(url: string) {
+	try {
+		return new URL(url)
+	} catch {
+		return null
+	}
+}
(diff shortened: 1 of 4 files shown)


nickvergessen (commented):
Not sure it should be there or in the Conversation+ menu?

Or maybe via the search field in the left sidebar?

ShGKme:
> Not sure it should be there or in the Conversation+ menu?

I'd prefer something that will also work for joining as a guest in the future. And if possible - more visible.

> Or maybe via the search field in the left sidebar?

This sounds nice but seems a bit hidden to the end user. You need to know that you can past a link to the search field. And still handle errors.

ShGKme:
> Can we do a component in spreed and use it with `@talk/src` in desktop?

I am unsure if I understand what component you want to add and where to use, but in general we should never directly use files from talk in desktop or vise versa. It should be a public API, like API to extend one app by another we have in other apps.

For example, we may have the link/component on the web all the time, but have slightly different handling or optional handler registration in the future when the multi account is supported.

PR #1569, opened 2025-12-05
chore(ci): completely disable ts:check

ShGKme:
### ☑️ Resolves

- Disable an entire CI job instead of disabling the result reporting

### DIFF
--- a/.github/workflows/lint-typescript.yml
+++ b/.github/workflows/lint-typescript.yml
@@ -4,7 +4,9 @@
 
 name: Lint TypeScript
 
-on: pull_request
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-typescript.yml
```diff
@@ -4,7 +4,9 @@
 
 name: Lint TypeScript
 
-on: pull_request
+# FIXME: temporary disable typecheck, until we can perform typecheck without checking Talk
```
Antreesy: There's one more FIXME on line 20

P.S. we can't just ignore spreed directory, right?

Antreesy (approved):

PR #1600, opened 2025-12-18
fix(linux): title bar is not transparent

ShGKme:
### ☑️ Resolves

- Apparently, `#00000000` casts to `0` int, which counts as "no value" and fallbacks to default

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


nickvergessen (approved):
Confirmed in local client on Ubuntu 25.04

Issue #976, opened 2024-12-10
Notifications could be clearer


kaffeeringe:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**
I can hear the notification but it took me some time to figure out, where I should look.

**Describe the solution you'd like**
Teams highlights the task in the taskbar if there is a new message and it shows a 1 in a red circle.
Nextcloud only shows a little * in the task. That doesn't really tell you "hey, look here"

**Describe alternatives you've considered**
Of course that might also be an advantage, if you don't want to be disturbed. Or if Talk isn't at the moment the center of your attention. But then you should set your status to "Do not disturb"...

**Additional context**
Maybe there could be a setting "high attention notifications" with red bubbles and "low attention notifications" with blue bubbles.

Antreesy:
Hi @kaffeeringe , thanks for your report!

Could you provide more information on OS used, where it is missing?

ATM I see a black indicator in Windows taskbar, and tray icon indicator is not implemented yet:
![Image](https://github.com/user-attachments/assets/aacd2504-9217-4245-bcdf-0ddbca881576)

For now that sounds like a duplicate of https://github.com/nextcloud/talk-desktop/issues/391

nickvergessen:
On linux there is none.

Issue #391, opened 2023-10-27
Add activity indicators to taskbar and system tray icons


amalg:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

The Win32/64 version of Desktop client does not have any indications on the tray icon for new message counts. Other messaging channels like Facebook Messenger, WhatsApp, Skype, etc. will all show some sort of indicator that something has happened in the application that the user should look at (receipt of a new message).

**Describe the solution you'd like**
- I would like to have the application icon displayed in the taskbar have an indicator as to the number of unread new messages in the client application.
- I would like the system tray icon to display a small contrast colored dot in the upper left or upper right area of the icon to indicate new activity in the client application the user should attend to.

**Describe alternatives you've considered**
Alternatively, instead of a new message counter, the taskbar icon of the application could simply display a colored dot that contrasts with the application icon color to indicate new activity the user should be made aware of.

**Additional context**

Example of Nextcloud Talk with new unread messages compared to other communication clients:

![image](https://github.com/nextcloud/talk-desktop/assets/725551/00b758d9-7d26-4e3f-855a-8d2aea199a8c)

git529:
this is a must have

vagner-dias:
+1

djaniol:
Really appreciate this feature. In general talk is not too bad for usage however without notification like blinking talk icon the whole idea with application for win doesn't have much sense in general in my opinion. 
Some activity indication is very appreciated.

Saueee:
This issue should be priority. Talk has already problems with notification.
I guess people download the dektop version to be notified correctly if something going on.
No point on having the dektop app if it has the same problem.

Saueee:
Well sorry let me rephrase. 
Talk has very few configuration options for notifications.

With problems i mean:
1. 30 sec polling seems too slow. Should be configurable without building source.
2. Browser Tab showing an Asterisk instead of blinking like everything else which is much harder to notice.
3. Incoming calls should have an constant sound/notification until dismissed or accepted (if channel not muted). 
    Currently if you missed the short sound you cant notice at all even if active in nextcloud. Could be configurable too.
[...]

Issue #291, opened 2023-08-11
Notifications: add badge dot to the system tray icon when there is a new or unread message.


ShGKme:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

**Is your feature request related to a problem? Please describe.**

There is a notification about a new message and taskbar icon badge. But when the window is closed and the notification is closed, it is not visible that some notification is missed.

**Describe the solution you'd like**

Add a badge dot to the tray system tray icon when there is a new or unread message. 

**Describe alternatives you've considered**

Currently, we have a badge about a new notification similar to `*` in the title on web-client. But it could be more useful to have a badge about unread mentions.

The difference is that the current badge appears only on a notification and disappears on window/tab activating. While usually such badges are shown when there are any unread mentions, no matter if the window is open or it is an old message.

**Additional context**

Discord example:

![image](https://github.com/nextcloud/talk-desktop/assets/25978914/766109fd-49b4-48c2-9e42-aab933f166e9)

ecker00:
> _There is a notification about a new message and taskbar icon badge._

I don't see this working at the moment in the Task Bar? Installed version from a few days ago.

Example, I have unread messages in Talk but no badge in the taskbar, compared to Discord and Mattermost.

![image](https://github.com/nextcloud/talk-desktop/assets/1331760/d534c2c2-3579-4632-ae73-f2171d401780)

benjamindahl:
Same issue in macOS. Would also love to see message count
![Bildschirmfoto 2023-11-01 um 11 30 29](https://github.com/nextcloud/talk-desktop/assets/2698024/d956d14b-36b2-4016-8b99-5aeabe11ba62)
![Bildschirmfoto 2023-11-01 um 11 30 38](https://github.com/nextcloud/talk-desktop/assets/2698024/ce32e4a0-6e3e-4e00-bb8f-d8175014bf8a)

ShGKme:
@xataxxx the original issue is a bit more complex. It is about showing a permanent marker depending on the number of all unread chats/mentions.

This badge is only used to notify the user about new notifications when the window is minimized (to the taskbar, not to the system tray).

[...]

PR #1605, opened 2025-12-21
feat: Add unread message badge to system tray icon

kupietz:
Shows numbered badge (1, 2, 15, 99+) on tray icon when there are unread conversations, respecting notification settings.

Features:
- Canvas-based badge rendering using offscreen BrowserWindow
- Respects notification settings per conversation
- Event-driven updates via store.subscribe() (no polling!)

### ☑️ Resolves

- #391
- #976 
- #291

### AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image  |  image


kupietz:
Successfully rebased on v2.2.0

Flofler:
Could someone please review this merge request? I’ve tested it successfully and it works perfectly. This is a must-have feature for my colleagues and me. We frequently overlook notifications because we don't want to keep the window open all the time—plus, the current setup doesn't work well when using multiple workspaces.

src/talk/renderer/TalkWrapper/useBadgeCountIntegration.ts
```diff
@@ -3,55 +3,103 @@
...
+	}
+	isInitialized = true
 
-	window.OCA.Talk.instance.$store.watch(countUnreadConversations, (newValue: number) => {
-		count.value = newValue
-	}, { immediate: true })
+	// Wait for store to be available, then subscribe to mutations
+	const checkAndSubscribe = () => {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		const store = (window as any).OCA?.Talk?.instance?.$store
```
Antreesy: First of all, thanks for the contribution!

I'm concerned, that this would not survive a planned migration to a different state manager.

There is already a public event introduced to provide these computations on web:
````
emit('talk:unread:updated', {
	conversations: number,
	messages: number,
	mentions: number,
	mentionsDirect: number,
})
````

kupietz: > First of all, thanks for the contribution!
> 
> I'm concerned, that this would not survive a planned migration to a different state manager.
> 
> There is already a public event introduced to provide these computations on web:
> 
> ```
> emit('talk:unread:updated', {
> 	conversations: number,
> 	messages: number,
> 	mentions: number,
> 	mentionsDirect: number,
> })
> ```

Thanks for the helpful hint and the review! Good point. I'll rework the integration to use the public `talk:unread:updated` (and rebase onto current main and fix the lint error while at it).

One small issue with the public event:  unlike in my original PR, the public event's counters don't take per-conversation notification settings into account, so muted conversations will count as well. This is not nice, but no big issue for me. I hope for everybody else, too (opinions welcome!).

That said, maybe we can extend the event's payload in a follow-up PR to spreed with an additional counter that respects notification settings.

Antreesy: Sure, we can adjust payload upstream as well to respect notification settings / count additionally.

But I would wait for @ShGKme to give a general opinion on the code change before implementation, so we and you don't spend more time going back and forth with reviews and adjustments =)

kupietz: Sounds good! Waiting for @ShGKme is fine, I've had some practice since December =)
The rework for using the public event is no big deal, it's mostly done already anyway.

kupietz:
Call me impatient, but I have now force-pushed the rebased and reworked integration that uses the public `talk:unread:updated` event =)

It is rebased onto current main, and the `relaunchApp` lint error from CI is fixed as well.

miaulalala:
@ShGKme are there any blockers to merge this?

miaulalala:
Hey @kupietz , can you add the right AI trailer to the commit? You can find the info on that in the AGENTS.md of this repo.

Antreesy:
CI might fail on forks, so don't worry. Assisted-by trailer looks correct

PR #1614, opened 2025-12-29
chore: migrate from @nextcloud/files deprecated API

ShGKme:


Antreesy (approved):
IIRC, there was only import path and functions name change, so should be good

Issue #1621, opened 2026-01-10
Updates about notifications are not opening


kikoso:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Open an outdated version of Nextcloud Talk (testing this on Mac 26.1 (25B78))
2. A notification pops up, pointing out to a new version available. It suggests to click on the notification, in order to update the app. 
3. Clicking on the notification does nothing.

<img width="330" height="81" alt="Image" src="https://github.com/user-attachments/assets/947acee4-d17a-438c-89a0-ed9149d43ba0" />

### Expected behaviour
Clicking on the notification should do what it says (opening the releases website)

### Actual behaviour
Clicking does nothing. 

## Diagnosis and logs

<details>
<summary>Diagnosis report</summary>

Logs from Nextcloud.

### Diagnosis report

| **Nextcloud Talk Desktop**      |                                                                                                                       |
| ------------------------------: | --------------------------------------------------------------------------------------------------------------------- |
| **Version**                     | v2.0.3                                                                                                                |
| **Built-in Talk version**       | v22.0.2                                                                                                               |
| **Release channel**             | stable                                                                                                                |
| **Operating system**            | Darwin 25.1.0 (Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000) |
| **Executable Path**             | /Applications/Nextcloud Talk.app/Contents/MacOS/Nextcloud Talk                                                        |
| ****Connected to****            | -                                                                                                                     |
| **Nextcloud version**           | 31.0.12                                                                                                               |
| **Nextcloud Talk version**      | 21.1.7                                                                                                                |
| **`notifications` app enabled** | ✅ yes                                                                                                                 |
| **`notify_push` app enabled**   | ❌ no                                                                                                                  |

#### Application config

```json
{
  "launchAtStartup": false,
  "theme": "default",
  "systemTitleBar": false,
  "monochromeTrayIcon": true,
  "zoomFactor": 1,
  "playSoundChat": "respect-dnd",
  "playSoundCall": "respect-dnd",
[...]

PR #1592, opened 2025-12-15
feat: notify about updates in main menu

max65482:
### ☑️ Resolves

- Fix: #155

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before (and OS notification) | after

When a new update is available, the burger menu gets a little notification dot and a new menu entry for updating appears. This replaces the often overlooked OS notification.

PR #1622, opened 2026-01-10
fix: prevent garbage collection of notification objects and ensure proper initialization order

kikoso:
### ☑️ Resolves

- Fix: #1621

#### 📝 Description
This PR fixes the issue where clicking on the "New version available" notification did nothing. 

The root cause was identified as the [Notification](cci:1://file:///Users/enriquelopezmanas/Documents/talk-desktop/src/app/githubReleaseNotification.service.js:156:0-165:1) object being garbage collected by Electron's main process before the user could interact with it. Additionally, the notification scheduler was starting before the app was fully ready.

**Changes:**
- **Persistent Notifications**: Introduced a `Set` to hold references to Notification objects in [githubReleaseNotification.service.js](https://github.com/nextcloud/talk-desktop/blob/main/src/app/githubReleaseNotification.service.js) and [downloads.ts](https://github.com/nextcloud/talk-desktop/blob/main/src/app/downloads.ts), preventing garbage collection until they are closed or clicked.
- **Improved Initialization**: Moved the update check scheduler inside `app.whenReady()` in [main.js](https://github.com/nextcloud/talk-desktop/blob/main/src/main.js).
- **Standardized URL**: Updated the release URL format to use the standardized `/tag/` prefix.

| 🏚️ Before | 🏡 After |
| :--- | :--- |
| Clicking the update notification does nothing | Clicking the notification opens the GitHub release page correctly |

### DIFF
--- a/src/app/downloads.ts
+++ b/src/app/downloads.ts
@@ -97,7 +97,7 @@ export function applyDownloadHandler(browserWindow: BrowserWindow) {
 			if (notification) {
 				notifications.add(notification)
 				notification.on('close', () => {
-					notifications.delete(notification!)
+					notifications.delete(notification)
 				})
 				notification.show()
 			}
--- a/src/app/githubReleaseNotification.service.js
+++ b/src/app/githubReleaseNotification.service.js
@@ -70,12 +70,6 @@ async function getLatestStableReleaseVersion() {
 	}
 }
 
-/**
- * Persistent set to store notification objects to prevent garbage collection
- *
- * @type {Set<Notification>}
- */
-const notifications = new Set()
 
 /**
  * Show native notification about the new version.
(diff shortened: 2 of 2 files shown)


src/app/downloads.ts
```diff
@@ -89,7 +94,13 @@ export function applyDownloadHandler(browserWindow: BrowserWindow) {
 				})
 			}
 
-			notification?.show()
+			if (notification) {
+				notifications.add(notification)
+				notification.on('close', () => {
+					notifications.delete(notification!)
+				})
```
ShGKme: One thing I'm worried about is that the `close` event is not guaranteed to be fired, which may lead to a small memory leak... But after checking a number of Electron issues, I don't see any better solution =\

src/app/githubReleaseNotification.service.js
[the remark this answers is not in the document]
kikoso: I've reverted the changes in src/app/githubReleaseNotification.service.js, keeping only the /tag/ fix in the URL as requested. This should resolve the overlap with PR #1592.

src/app/downloads.ts
kikoso: The Set is used to prevent the Notification object from being garbage collected by Electron's main process while the OS is still displaying it. This is a known behavior where if no JavaScript reference is kept, the click event listeners are destroyed before they can be triggered. I am not sure if there is a better solution, either - but Download notifications are infrequent, so I guess this will not kill the app through memory leaks.

kikoso:
@ShGKme , rebased and committed to fix the DCO.

PR #1641, opened 2026-01-29
ci: fix version comparison with update-built-in-talk on major updates

ShGKme:
### ☑️ Resolves

- Ref: https://github.com/nextcloud/talk-desktop/pull/1638
- `parseInt('v23')` is `NaN`
- Simplify comparison by using `localCompare` with `numeric`


nickvergessen (approved):
Writing tests (or cases) is an option 🙊

PR #1662, opened 2026-02-12
fix(welcome): add brandGradient build config for Welcome screen

ShGKme:
### ☑️ Resolves

- Styling the welcome screen as the logo

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### DIFF
--- a/src/welcome/welcome.html
+++ b/src/welcome/welcome.html
@@ -23,6 +23,10 @@
 		-webkit-app-region: drag;
 	}
 
+	body {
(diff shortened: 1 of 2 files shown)


src/welcome/welcome.html
```diff
@@ -23,6 +23,10 @@
 		-webkit-app-region: drag;
 	}
 
+	body {
+		background: <%= __BUILD_CONFIG__.brandGradient %>;
```
Antreesy: Doesn't it need any dependencies? Or built-in with Electron?

PR #1666, opened 2026-02-19
feat: download installer on "Update" click instead of opening the GitHub release web-page

ShGKme:
### ☑️ Resolves

- A user won't need to manually find the correct installer and download it on the scary GitHub release page
- If in the future the current installer isn't available anymore — fallback to opening the release page
- Next steps:
  - New section in the settings to select the release channel and have more info about the update (requires new strings)
  - Auto-update
- Tested on all Linux and Windows installers + fallback case
 
### 🖼️ Screenshots

https://github.com/user-attachments/assets/83d86990-02ab-46b6-ad8f-752efe26f08d

### DIFF
--- a/src/app/githubRelease.service.ts
+++ b/src/app/githubRelease.service.ts
@@ -4,71 +4,105 @@
  */
 
 import { BrowserWindow } from 'electron'
-import semver from 'semver'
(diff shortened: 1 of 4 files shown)


Antreesy:
NVM, tested the wrong branch

PR #1672, opened 2026-02-25
fix(linux/flatpak): ensure emojis are rendered correctly on every launch

ShGKme:
### ☑️ Resolves

- We added a migration fixing Flatpak emoji rendering in https://github.com/nextcloud/talk-desktop/pull/1636
- But it may be broken again, for example, on Ubuntu update...
- This PR adds a check that:
  - Renders emoji on a canvas
  - Checks if it is colored
  - If it is not, clears the font config cache again
  - Saves the last check timestamp to not fall into an infinite loop if clearing doesn't help anymore


Antreesy (commented):
Looks fine, but I can't test and not sure it won't give any false-positives, and won't change in the future (the way we can catch it by some sort of test covering)

ShGKme:
> it won't give any false-positives

`Noto Color Emoji` is quite a bright font and bundled in Flatpak.

It may only be false positive if a user has a custom config with a custom emoji font that is very grayscale. But even in this case, it quickly relaunches the app and continues.

An alternative would be an explicit UI setting:
- https://github.com/nextcloud/talk-desktop/pull/1637

@nickvergessen What do you think?

ShGKme:
Waiting for more users confirming they have this issue again.

PR #1741, opened 2026-05-14
build: `.overrides/` for build customization

ShGKme:
### 1. Refactoring

Rename `config` to `build config` where relevant for consistency and distinguishing from `app config`

### 2. Build customization with `.overrides/`

**Before**

Customizing using the build config file with any name located on `env.CUSTOM_CONFIG` path.

**After**

Customizing using fixed and ignored `.overrides/build.config.json`

**Motivation**

- Predictable customization — the file is always on the same ignored location
- Follow-up features — storing custom theming and icons in the same `.overrides/` instead of changing files in place


ShGKme:
- Also updated the name in REUSE

nickvergessen (approved):

PR #1745, opened 2026-05-20
feat: add custom CSS styling

TimonBed:
Adds a local Custom CSS option to Talk Desktop settings using the existing app config storage. Changes apply immediately in the desktop renderer, and clearing the field removes the custom styling. We use custom CSS internally, for example to display ticket messages from our ticket system and for theming.

### DIFF
--- a/src/app/AppConfig.ts
+++ b/src/app/AppConfig.ts
@@ -6,7 +6,7 @@
 import { app, webContents } from 'electron'
 import { readFile, writeFile } from 'node:fs/promises'
 import { join } from 'node:path'
-import { isWayland, isMac } from './system.utils.ts'
+import { isMac, isWayland } from './system.utils.ts'
 
 const APP_CONFIG_FILE_NAME = 'config.json'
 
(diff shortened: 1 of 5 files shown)


Antreesy:
Hi, thanks for the contribution!
We haven't replied yet, since we're still discussing the options internally. Will get back to you soon

ShGKme:
Hello Timon,

Thank you for your contribution.

We are discussing if this feature makes sense as a user feature. Adding custom CSS is always risky. This is something not guaranteed to work across upgrades, and it may break the UI.

If we add it, it shouldn't be so prominent for a general user. This is not something an average user would do on the same level as switching between a light/dark color scheme. Especially if it is a text area to manually input the CSS code.

At the same time, for a sysadmin manually opening the client on every machine to insert the CSS may not be the most convenient solution.

Instead of storing the CSS code in the config, I'd use a CSS file. For example, a fixed `custom.css` file in the config location (`app.getPath('userData')`). Then a sysadmin can just add the file automatically on installation. Or in `/ProgramData` for multi-user setups.

In the app settings then it should be indicated that the custom CSS is applied.

---

An alternative way would be to include the styles via `.overrides` during the custom client build. This would allow distributing the client with built-in styles. But you'd need to build the client for your users locally. @TimonBed would this work for you?

TimonBed:
ty for the feedback.

Maintaining a custom build is something we would like to move away from in the future, so the .overrides approach would not be ideal for us long-term.

I reworked the PR to avoid exposing custom CSS as a regular user-facing setting. It now loads a fixed custom.css file from the app data location, with support for %ProgramData%\Nextcloud Talk\custom.css on Windows for managed/multi-user deployments. The settings UI only indicates when custom CSS is applied.

After custom css file added. Including indicator
Screenshot 2026-05-29 113236

PR #1747, opened 2026-05-21
feat: add enterprise server URL configuration

TimonBed:
Add enterprise policy support so IT admins can preconfigure the Nextcloud server URL via Windows GPO (HKLM/HKCU\Software\Policies\Nextcloud\Talk) or macOS MDM preferences on com.nextcloud.talk.mac. The app reads ServerUrl at startup and optionally EnforceServerUrl to prefill the login field and skip the setup window when enforcement is enabled.

### DIFF
--- a/src/app/policy.config.ts
+++ b/src/app/policy.config.ts
@@ -9,6 +9,7 @@ import { isMac, isWindows } from './system.utils.ts'
 
 export type PolicyConfig = {
 	serverUrl?: string
+	enforceServerUrl?: boolean
 }
 
 /**
@@ -17,7 +18,7 @@ export type PolicyConfig = {
  * @param key - Registry key
  * @param name - Value name
  */
-function getWindowsPolicyValue(key: string, name: string): string | undefined {
+function getRegistryValue(key: string, name: string): string | undefined {
 	try {
 		const stdout = execFileSync('reg', ['query', key, '/v', name], { encoding: 'utf8', windowsHide: true })
 		return stdout.match(new RegExp(`\\s${name}\\s+REG_\\w+\\s+(.+)`, 'i'))?.[1]?.trim()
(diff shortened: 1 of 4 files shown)


src/authentication/renderer/AuthenticationApp.vue
```diff
@@ -29,7 +29,7 @@ let [prefilledServer, prefilledUser] = atIndex === -1
 	: [prefilledAccount.slice(atIndex + 1), prefilledAccount.slice(0, atIndex)]
 
 const rawServerUrl = ref(BUILD_CONFIG.domain ?? prefilledServer)
-const enforceDomain = Boolean(BUILD_CONFIG.domain && BUILD_CONFIG.enforceDomain)
+let enforceDomain = Boolean(BUILD_CONFIG.domain && BUILD_CONFIG.enforceDomain)
```
ShGKme, line 32: This makes it a general, none-reactive but dynamic variable. Vue won't react to its change.

src/app/policy.config.ts
```diff
@@ -0,0 +1,58 @@
...
+		return undefined
+	}
+}
+
+/**
+ * Get deployment configuration from OS policy.
+ */
+export function getPolicyConfig(): PolicyConfig {
+	if (isWindows) {
+		for (const key of ['HKLM\\Software\\Policies\\Nextcloud\\Talk', 'HKCU\\Software\\Policies\\Nextcloud\\Talk']) {
```
ShGKme, line 35: The default vendor is `Nextcloud GmbH`, and it can be different in a custom build.
Both the vendor name and the application name are available in the `BUILD_CONFIG`.

src/app/policy.config.ts
```diff
@@ -0,0 +1,58 @@
...
+			const serverUrl = getRegistryValue(key, 'ServerUrl')
```
ShGKme, line 36: The current predefined server URLs value is a bit more complicated. We are using the `accounts` config which includes a list of the servers (for a multi account in the future) and the userids (to predefine them as well). I'd need to think a bit on how to name the config, but it would be good to support the same features we support via the existing config and the command line command.

src/authentication/renderer/AuthenticationApp.vue
```diff
@@ -44,7 +44,13 @@ const serverUrl = computed(() => {
 const state = ref('idle')
 const stateText = ref('')
 
-onMounted(() => {
+onMounted(async () => {
+	const { serverUrl, enforceServerUrl } = await window.TALK_DESKTOP.getPolicyConfig()
+	if (serverUrl) {
+		enforceDomain ||= Boolean(enforceServerUrl)
+		rawServerUrl.value = serverUrl
+	}
```
ShGKme, line 52: This moves the default (predefined) configuration handling from the config module to the UI. It should be on the configuration itself (the app config). Otherwise the app config used in another place has a different value.

ShGKme (changes requested):
Hi Timon,

This PR adds 2 features:
- Predefined server URL that can also be set via command line command by the user and is defined by the last used server. This is a general user feature, managed by the user that can be improved with GPO
- Enforced server URL that is not available for the end user in any way. It targets only enterprise installations and duplicates built-in setting (the build config).

Though they seem related, they are targeting different flows and areas.

Could you keep in this PR only the first feature?

---

I can see, your PR is generated with an AI agent. We actually plan to have this feature in the next minor release (this week), and also are working on this feature. 

Could you tell us if you want to continue working on your PR and have time for it in the nearest days? Or we can take over?

To make sure we have this feature in before the release (and to not use a PR as middleware between the reviewers and an AI-agent 👀).

TimonBed:
> Hi Timon,
> 
> This PR adds 2 features:
> 
> * Predefined server URL that can also be set via command line command by the user and is defined by the last used server. This is a general user feature, managed by the user that can be improved with GPO
> * Enforced server URL that is not available for the end user in any way. It targets only enterprise installations and duplicates built-in setting (the build config).
> 
> Though they seem related, they are targeting different flows and areas.
> 
> Could you keep in this PR only the first feature?
> 
> I can see, your PR is generated with an AI agent. We actually plan to have this feature in the next minor release (this week), and also are working on this feature.
> 
> Could you tell us if you want to continue working on your PR and have time for it in the nearest days? Or we can take over?
> 
> To make sure we have this feature in before the release (and to not use a PR as middleware between the reviewers and an AI-agent 👀).

Hi,

Since I'm only in the office for a few more hours today and won't be able to test the GPO/MDM configs over the weekend, please feel free to take over the PR. I don't want to hold up your release timeline for this week.

I'll pause working on this on my end. Once your changes are merged, just let me know. I will test the behavior on Monday and check if it works for us.

Have a great weekend

Best,
Timon

ShGKme:
Hey @TimonBed, thanks again for your contribution.

Sorry we had a delay. I'm closing your PR because the feature has been implemented in:
- https://github.com/nextcloud/talk-desktop/pull/1797
- https://github.com/nextcloud/talk-desktop/pull/1800

The main differences are:
- Using Windows API via prebuilt binary in our for instead of `reg.exe` which might not be available in a strict corporate environment
- Support `accounts` option as in CLI for future support of userId and multi-account
- The managed value is read on the application config level, not from theUI request
- No `enforceServerUrl` since it reflects the build configuration rather than application config and not user-controlled unlike the prefilled values

And thank you for the `systemPreferences.getUserDefault`. Before seeing it in your code, I didn't notice Electron had a built-in API for reading the managed configuration on macOS (I wish they also had it for Windows).

Thank you for your contribution again. We will re-review and merge the second one with the custom CSS soon after the release (so translators have time before the next release).

PR #1756, opened 2026-06-08
feat(Viewer): add zoom/pan and rotate to image viewer

### LINKED ISSUE #1535: macOS: Zoom on pictures
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.

---

In the browser version you are able to zoom in and out to see pictures

It would be great to be able to zoom in on pictures from within the talk client app

### LINKED ISSUE #1757: Image viewer: add rotate left/right buttons

When viewing an image sent in a Talk chat, there is no way to rotate it if it was taken in the wrong orientation (e.g. a portrait photo opened as landscape). The viewer currently has no rotation controls.

Add rotate left (−90°) and rotate right (+90°) buttons to the image viewer actions menu. The rotation should be visual only (display transform), not modifying the original file. Rotating works in combination with zoom and pan.

- Editing the image outside of Talk and re-uploading — inconvenient for quick viewing
- Using "Open in web browser" to view in Files — breaks the in-app experience

**Additional context**

The Talk Desktop viewer (`ViewerHandlerImages.vue`) renders images locally via Electron, so rotation can be applied purely as a CSS `transform` without server-side changes. A working implementation is proposed in the linked PR.

vladopol:
## Summary

Adds zoom/pan and rotate controls to the image viewer in Talk Desktop.

### Zoom and pan

Installs [`panzoom`](https://github.com/anvaka/panzoom) (v9.4.4, 6 kB) — same library already used in spreed for screen share zoom ([spreed#14028](https://github.com/nextcloud/spreed/pull/14028)).

- Scroll wheel / trackpad pinch → zoom in/out (1× – 8×), centered on cursor
- Drag to pan when zoomed in
- Double-click to zoom in 3× at cursor position; double-click again to reset to 1×
- Auto-centers image when zoom returns to 1× (smooth animation)
- Panning blocked at minimum zoom to prevent accidental image shifts
- Cursor feedback: `zoom-in` at 1×, `grab` / `grabbing` when zoomed

Addresses the macOS limitation where pinch-to-zoom had no effect and Cmd+/- zoomed the entire application UI instead of the image.

Closes #1535

### Rotate

Adds rotate left (−90°) and rotate right (+90°) buttons to the actions menu.

- Rotation state and controls are fully self-contained in `ViewerHandlerImages`
- Unbounded angle increments ensure the CSS transition always animates in the correct direction
- Rotation resets when switching images
- `ViewerHandlerImages` exposes an `actions` array; `ViewerApp` renders it generically via `component :is` — no image-specific code leaks into the generic component

Closes #1757

## Architecture

Handler components can expose an `actions` array to contribute entries to the viewer's action menu without modifying `ViewerApp`. Each action is `{ key, label, icon, onClick }`. This keeps `ViewerApp` generic and all feature-specific logic inside the handler.

## Test plan

- [ ] Open an image in the viewer
- [ ] Scroll wheel / trackpad pinch zooms image centered on cursor
- [ ] Drag image while zoomed in — panning works, cursor shows grab/grabbing
- [ ] Double-click to zoom in 3×; double-click again to reset
- [ ] Open a second image — zoom/pan state resets
- [ ] Actions menu shows Rotate left, Rotate right, Open in web browser
- [ ] Rotate right repeatedly — image turns clockwise in 90° steps with smooth animation
- [ ] Rotate left repeatedly — counter-clockwise
- [ ] Mix left and right — always rotates in the correct direction
- [ ] Switch to another image — rotation resets to 0°
- [ ] Open a video in the viewer — rotate buttons are not shown
- [ ] Zoom in, then rotate — both transforms apply independently

Assisted-by: ClaudeCode:claude-sonnet-4-6

### DIFF
--- a/src/talk/renderer/Viewer/ViewerHandlerImages.vue
+++ b/src/talk/renderer/Viewer/ViewerHandlerImages.vue
@@ -68,16 +68,25 @@ function disposePanzoom() {
 	grabbing.value = false
 }
 
+/**
+ * @param {Function} handleLoadEnd - Callback to signal load completion
+ */
 function onImageLoad(handleLoadEnd) {
 	handleLoadEnd(false)
 	initPanzoom()
 }
 
+/**
+ * @param {Function} handleLoadEnd - Callback to signal load error
+ */
 function onImageError(handleLoadEnd) {
 	handleLoadEnd(true)
 	disposePanzoom()
 }
 
+/**
+ * @param {MouseEvent} event - The double-click event
+ */
 function onDoubleClick(event) {
 	if (!instance.value) {
 		return
--- a/src/talk/renderer/Viewer/ViewerHandlerImages.vue
+++ b/src/talk/renderer/Viewer/ViewerHandlerImages.vue
@@ -125,7 +125,7 @@ function onDoubleClick(event) {
 	if (scale.value === 1) {
 		instance.smoothZoom(x, y, ZOOM_FACTOR)
 	} else {
-		instance.smoothZoomAbs(x, y, 0)
+		instance.smoothZoomAbs(x, y, ZOOM_MIN)
 	}
 }
 
--- a/src/talk/renderer/Viewer/ViewerHandlerImages.vue
+++ b/src/talk/renderer/Viewer/ViewerHandlerImages.vue
@@ -4,7 +4,8 @@
 -->
 
 <script setup>
-import { computed } from 'vue'
+import panzoom from 'panzoom'
+import { computed, onBeforeUnmount, ref, watch } from 'vue'
 import ViewerHandlerMedia from './ViewerHandlerMedia.vue'
 import { generateFilePreviewUrl } from './viewer.utils.ts'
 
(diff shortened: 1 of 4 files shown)


nickvergessen:
You need to sign-off your commits otherwise we can not merge them at all

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,135 @@ const props = defineProps({
 		required: true,
 	},
 })
+const ZOOM_MIN = 1
+const ZOOM_MAX = 8
+const ZOOM_FACTOR = 3
```
ShGKme, line 23: Thakn you for adding the constants

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,135 @@ const props = defineProps({
...
+	if (scale.value === 1) {
+		return 'viewer-image--zoom-in'
+	}
+	return grabbing.value ? 'viewer-image--grabbing' : 'viewer-image--grab'
+})
+
+function initPanzoom() {
+	if (!wrapperRef.value) {
+		return
+	}
```
ShGKme, line 57: **nitpick(non-blocking):** if panzoom is only initiated on the image load, `wrapperRef` is definitely defined. If it isn't (e.g. due to a bug), it is better to fail here and have an error rather than exit silently.

```suggestion
```

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,135 @@ const props = defineProps({
...
 	},
 })
+const ZOOM_MIN = 1
+const ZOOM_MAX = 8
+const ZOOM_FACTOR = 3
 
 const src = computed(() => generateFilePreviewUrl(props.file.fileid, props.file.etag))
+
+const wrapperRef = ref(null)
+const instance = ref(null)
```
ShGKme: This variable is never used as a reactive state

```suggestion
let instance = null
```

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,135 @@ const props = defineProps({
 		required: true,
 	},
 })
+const ZOOM_MIN = 1
+const ZOOM_MAX = 8
+const ZOOM_FACTOR = 3
 
 const src = computed(() => generateFilePreviewUrl(props.file.fileid, props.file.etag))
+
+const wrapperRef = ref(null)
```
ShGKme: Since 3.5 Vue has a more advanced way to define a template ref. 
https://vuejs.org/api/composition-api-helpers.html#usetemplateref

```suggestion
const panzoomWrapperRef = useTemplateRef('panzoomWrapper')
```

vladopol:
Fixed the JSDoc lint errors (`jsdoc/require-param-type`) in the latest commit. Could you approve the CI run?

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,144 @@ const props = defineProps({
...
-		<img
-			:key="src"
-			class="viewer-image"
-			:class="mediaClass"
-			:src="src"
-			:alt="file.basename"
-			@load="handleLoadEnd(false)"
-			@error="handleLoadEnd(true)">
+	<ViewerHandlerMedia v-slot="{ handleLoadEnd }">
+		<div class="viewer-image-container" @dblclick.capture="onDoubleClick">
```
ShGKme, line 163: **question:** Why does it need the `capture` mode (and `preventDefault` + `stopPropagation` in the handler)?

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,144 @@ const props = defineProps({
...
+	cursor: zoom-in;
+}
+
+.viewer-image--grab {
+	cursor: grab;
+}
+
+.viewer-image--grabbing {
+	cursor: grabbing;
+}
```
ShGKme, line 210: **issue(blocking):** `cursor` is not inheritable in CSS. Because the styles are applied on the container, the image itself always has the `default` cursor.

src/talk/renderer/Viewer/ViewerHandlerImages.vue
ShGKme, line 1: **issue(blocking)**: currently it is possible to move the image even on `scale: 1`, moving it almost completely away:
Initial | Moved
--- | ---
Image | Image

It should only be movable on the axis, if the image on this axis eexceeds its container limits.

ShGKme (commented):
Thank you for your contribution. All good in general, but there are some problems.

---

We have a new [AI Policy](https://github.com/nextcloud/.github/blob/master/AI_POLICY.md). If this was created with AI, could you mention it in the PR description and add the `Assisted-by:` git trailer?

src/talk/renderer/Viewer/ViewerHandlerImages.vue
vladopol: The `.capture` modifier is needed because panzoom registers its own `dblclick` listener on the same element — without `capture`, both handlers fire and the image zooms twice (once by panzoom's built-in handler, once by ours). Intercepting in the capture phase lets us handle it first and call `stopPropagation` to prevent panzoom's handler from running.

`preventDefault` suppresses the browser's default double-click text selection behavior. `stopPropagation` also prevents the event from bubbling up to the parent Viewer, which closes on double-click.

src/talk/renderer/Viewer/ViewerHandlerImages.vue
ShGKme: It is fixed now for `scale: 1`, but zoomed it is possible to move the image completely out of the viewport.

ShGKme:
There is still the same problem for me:

https://github.com/user-attachments/assets/f5ef4082-8662-46ca-8a79-b68035679aeb

ShGKme:
Pan just gets stuck when the original image (not the container) is reached.

https://github.com/user-attachments/assets/4db6c878-d5c8-4f0e-ac43-c76951dcdcf3

src/talk/renderer/Viewer/ViewerHandlerImages.vue
ShGKme: I see. Indeed, panzoom doesn't allow customizing the double-click behavior directly via the panzoom options...

One nitpick: `preventDefault` and `stopPropagation` can also be set via Vue like `capture` mode is, to not mix Vue and JS ways for the same event:
```
@dblclick.capture.stop.prevent="onDoubleClick"
```

ShGKme:
Also, zooming out with double-click may result in the image being completely disappeared depending on the double-click position...

ShGKme:
After zooming out on the second double click, the image is stuck in some uneven position, and neither pan nor zoom works again.

problem

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +18,194 @@ const props = defineProps({
...
+	})
+	instance.on('panstart', (pz) => {
+		if (pz.getTransform().scale <= ZOOM_MIN) {
+			return
+		}
+		grabbing.value = true
+	})
+	instance.on('pan', () => {
+		clampToBounds()
+	})
```
Antreesy, line 88: This triggers infinite loop after 1-2 user drag interactions only for me?

function `clampToBounds()` looks like it tries to replicate native panzoom options `bounds: true` and `boundsPadding: 1`, why wouldn't use them directly?

ShGKme:
Hi @vladopol,

After several rounds of reviews, the implemented feature doesn't work correctly in its basic functionality. The issues could have been found with quick manual testing. This suggests the changes were AI-generated and submitted without being verified first. This violates our [AI policy](https://github.com/nextcloud/.github/blob/master/AI_POLICY.md) in **author accountability** and **code quality** requirement. Therefore, the PR will be closed according to the **maintainer's discretion**.

Thanks for your interest in contributing to our project. Feel free to open PR again when the feature is ready and tested. Or we may implement it later ourselves.

vladopol:
Reworked and resubmitted as #1812. The review findings were valid. Details, the fix, and how each of your scenarios is now covered by automated and manual tests are in the new PR. Thanks for the thorough review and the feedback comments — they're really valuable.

PR #1758, opened 2026-06-08
feat(Viewer): add rotate left/right to image viewer

### LINKED ISSUE #1757: Image viewer: add rotate left/right buttons

vladopol:
## Summary

Adds rotate left and rotate right controls to the image viewer actions menu, so users can correct the orientation of images received in Talk chats without leaving the app.

**Changes:**
- `ViewerHandlerImages.vue` — rotation state; `rotateLeft`/`rotateRight` exposed via `defineExpose`; CSS `transform: rotate()` applied to `` with a 0.3s transition; rotation resets when switching images
- `ViewerApp.vue` — two `NcActionButton` entries wired to `viewerRef.rotateLeft/Right`; buttons are shown only for handlers that expose these methods, so video and other viewers are unaffected

Rotation uses unbounded angle increments (`+90`, `+180`, ...) so the CSS transition always animates in the correct direction regardless of how many times the button is pressed.

Depends on #1756 (zoom/pan).

## Test plan

- [ ] Open an image in the viewer — actions menu shows Rotate left, Rotate right, Open in browser
- [ ] Rotate right repeatedly — image turns clockwise in 90° steps with smooth animation, direction never reverses
- [ ] Rotate left repeatedly — same counter-clockwise
- [ ] Mix left and right — always rotates in the correct direction
- [ ] Switch to another image — rotation resets to 0°
- [ ] Open a video in the viewer — rotate buttons are not shown
- [ ] Zoom in, then rotate — both transforms apply independently

Fixes #1757

### DIFF
--- a/src/talk/renderer/Viewer/ViewerApp.vue
+++ b/src/talk/renderer/Viewer/ViewerApp.vue
@@ -7,9 +7,12 @@
 import { translate as t } from '@nextcloud/l10n'
 import { generateUrl } from '@nextcloud/router'
 import { computed, ref } from 'vue'
+import NcActionButton from '@nextcloud/vue/components/NcActionButton'
 import NcActionLink from '@nextcloud/vue/components/NcActionLink'
 import NcModal from '@nextcloud/vue/components/NcModal'
 import IconOpenInNew from 'vue-material-design-icons/OpenInNew.vue'
+import IconRotateLeft from 'vue-material-design-icons/RotateLeft.vue'
+import IconRotateRight from 'vue-material-design-icons/RotateRight.vue'
 
 /**
  * Noop
@@ -19,6 +22,7 @@ function noop() {}
 const isOpen = ref(false)
 const onClose = ref(noop)
 const file = ref(null)
+const viewerRef = ref(null)
 
 const viewComponent = computed(() => file.value && window.OCA.Viewer.availableHandlers.find((handler) => handler.mimes.includes(file.value.mime))?.component)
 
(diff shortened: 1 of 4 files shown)


src/talk/renderer/Viewer/ViewerApp.vue
```diff
@@ -67,9 +71,22 @@ defineExpose({
...
+					<IconRotateLeft :size="20" />
+				</template>
+				{{ t('talk_desktop', 'Rotate left') }}
+			</NcActionButton>
+			<NcActionButton v-if="viewerRef?.rotateRight" @click="viewerRef.rotateRight()">
+				<template #icon>
+					<IconRotateRight :size="20" />
+				</template>
+				{{ t('talk_desktop', 'Rotate right') }}
+			</NcActionButton>
```
ShGKme, line 89: **issue(blocking)**: This component is a generic viewer component and shouldn't know anything about a specific view - image viewer and its operations.

Instead of conditionally rendering buttons here, we can add a slot, for example, `#actions`, and pass the necessary buttons directly in the `ViewerHandlerImages`. Exposing public methods also won't be necessary in this case.

src/talk/renderer/Viewer/ViewerHandlerImages.vue
```diff
@@ -14,19 +15,153 @@ const props = defineProps({
...
+	height: 100%;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+}
+
+.viewer-image {
+	max-width: 100%;
+	max-height: 100%;
+	transition: transform 0.3s ease;
```
ShGKme, line 153: **issue(blocking)**: to use the theming variable:
```suggestion
	transition: transform var(--animation-slow) ease;
```

ShGKme (commented):
Thank you, I think we can add it. I'd only refactor this part a little bit to avoid extending a generic component with image-specific features, keeping everything in the image viewer.


---

> Depends on https://github.com/nextcloud/talk-desktop/pull/1756 (zoom/pan).

Thank you for initially splitting it into 2 PRs. Though, these 2 features are related. To simplify it, we can merge it in a single PR `feat(Viewer): add image panzoom and rotation`

---

We have a new [AI Policy](https://github.com/nextcloud/.github/blob/master/AI_POLICY.md). If this was created with AI, could you mention it in the PR description and add the `Assisted-by:` git trailer?

vladopol:
Closing in favour of #1756 which now includes both zoom/pan and rotate, with the refactored actions pattern suggested by @ShGKme.

PR #1793, opened 2026-07-03
refactor: remove `@nextcloud/initial-state@2` usage

ShGKme:
### ☑️ Resolves

- Currently, the initial state is used only for `notifications` app settings (like in the original app)
- App **mutates** initial state on settings and user status change
- Mutating the initial state is not compatible with the latest `@nextcloud/initial-state` 
- Instead of sharing the state via initial state, computing it on the only place it is actually needed.

For testing:
1. Enable dev mode
2. Open any conversation
3. Use Show Call Popup and listen for the sound depending on settings and user status

image


Antreesy (approved):
Tested against stable33

PR #1794, opened 2026-07-03
chore(eslint): migrate to ESLint 10 with @nextcloud/eslint-config@9.0.0

ShGKme:


Antreesy (approved):
Apart from weird issue with deprecated props rule, all good

PR #1799, opened 2026-07-08
build(scripts): only override styles on release if missing or outdated

ShGKme:
### ☑️ Resolves

- `build(server-styles): check if overrides are up-to-date`
  - When overrides are required, they must be on the same styles version and on the current build config's theming
- `refactor(scripts): split override-nextcloud-styles into CLI and function`
  - To call a function from CLI instead of one CLI from another CLI
- `build(override-nextcloud-styles): only override if missing or outdated`
  - If overridden styles are up-to-date, no need to extract them again, including on release

### DIFF
--- a/scripts/override-nextcloud-styles.mjs
+++ b/scripts/override-nextcloud-styles.mjs
@@ -3,10 +3,8 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-import { resolve } from 'node:path'
 import { argv, echo } from 'zx'
-import { getNextcloudVersionForTalk, resolveBuildConfig } from '../build/resolveBuildConfig.js'
-import { extractNextcloudStyles } from './nextcloud-app-host/extractNextcloudStyles.mjs'
+import { overrideNextcloudStyles } from './overrideNextcloudStyles.commands.mjs'
 
 if (argv.help) {
 	echo`Override Nextcloud styles with the current build config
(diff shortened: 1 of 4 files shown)


scripts/override-nextcloud-styles.mjs
```diff
@@ -17,7 +15,6 @@ if (argv.help) {
 	Arguments:
 		--version <nextcloud-version-major> - major version of Nextcloud to override styles for (e.g. "34"), default: the current built-in Talk's Nextcloud version
 		--help - show help
-		--port <number> - port to expose the Nextcloud server on, default: 6123
```
Antreesy, line 20: port is not needed anymore?

Antreesy (approved):
Code-wise makes sense to me

scripts/override-nextcloud-styles.mjs
ShGKme: No. In the past, the docker container with Nextcloud server had port exposed to fetch the styles on the host. To run 2 containers with different Nextcloud servers, we needed to set different ports.

Later, to avoid this issue entirely, we stopped exposing the port and switched to fetching styles via `docker exec curl`.

ShGKme: The `extract-nextcloud-styles` script doesn't have port anymore, so `--port` in `override-nextcloud-styles` does nothing.

PR #1809, opened 2026-07-12
fix(windows): make Windows ARM64 build start and package correctly

aequivalent2:
## Summary

One fixes needed for a working win32-arm64 build/release, on top of the packaging groundwork
from #1725:

1. **App crashes immediately on launch on arm64.** `getWindowsRegistryItem()`
   (`src/app/windows.utils.ts`) threw unconditionally on any non-x64 architecture, before ever
   attempting to load the native `@vscode/windows-registry` module. Since
   `readManagedConfig()` (`src/app/managedConfig.service.ts`) is called at module scope in
   `AppConfig.ts`, i.e. on every app start, this throw is never caught: Error: getWindowsRegistryItem is only available on x64

## Changes

- Resolve the native Registry module path per `process.arch` instead of a hardcoded
  `win32-x64` path; treat a missing prebuild for the current architecture as a recoverable
  error (return `undefined`) instead of throwing.

## Testing

- Applied and verified both patches against a fresh clone of `main` (`git am`, no conflicts).
- Verified `windows.utils.ts` change with esbuild (syntax/bundle check).

## Related

- #1725 (Windows ARM64 packaging support)

### DIFF
--- a/src/app/windows.utils.ts
+++ b/src/app/windows.utils.ts
@@ -19,14 +19,20 @@ export function getWindowsRegistryItem(hive: HKEY, path: string, name: string): 
 		throw new Error('getWindowsRegistryItem is only available on Windows')
 	}
 
-	if (process.arch !== 'x64') {
-		throw new Error('getWindowsRegistryItem is only available on x64')
+	if (process.arch !== 'x64' && process.arch !== 'arm64') {
+		throw new Error('getWindowsRegistryItem is only available on x64 and arm64')
+	}
+
+	let windowRegistry
+	try {
+		const require = createRequire(import.meta.url)
+		windowRegistry = require(`@vscode/windows-registry/prebuilds/win32-${process.arch}/@vscode+windows-registry.node`)
+	} catch {
+		console.error('Error loading @vscode/windows-registry')
+		return undefined
 	}
 
 	try {
-		const require = createRequire(import.meta.url)
-		const windowRegistry = require('@vscode/windows-registry/prebuilds/win32-x64/@vscode+windows-registry.node')
-
 		return windowRegistry.GetStringRegKey(hive, path, name)
 	} catch (error) {
 		// 'Unable to open registry key' is expected when the registry key does not exist


forge.config.js
```diff
@@ -244,7 +244,10 @@ module.exports = {
 			manufacturer: BUILD_CONFIG.companyName,
 			shortName: BUILD_CONFIG.applicationNameSanitized,
 			upgradeCode: BUILD_CONFIG.winUpgradeCode,
-			arch: 'x64', // electron-wix-msi defaults to x86 and arm64 is not supported
+			// electron-wix-msi (>=5.1.3, in use here) supports 'x86' | 'x64' | 'arm64' as Platform
+			// via WiX Toolset >=3.14's candle.exe. Use the actual packaging target instead of a
+			// hardcoded 'x64', otherwise win32-arm64 builds ship an MSI mislabeled as x64.
+			arch: TARGET_ARCH,
```
ShGKme: The comment isn't entirely true. WiX Toolset >=3.14's `candle.exe` indeed claims to support arm64.

However, `electron-wix-msi@5.1.3` doesn't. It is documented as 

> Values can be either `x86` or `x64`

https://github.com/electron-userland/electron-wix-msi/blob/v5.1.3/README.md?plain=1#L150-L151

And the source code checks for this property like it is either one or another.

For example: https://github.com/electron-userland/electron-wix-msi/blob/f62348b06b8dd855a13f276782d862f3acc16bd2/src/creator.ts#L410

ShGKme: Probably, we can use `arm64` here. But I'm unsure if it is needed. This is only the installer arch, not the installed app, isn't it?

ShGKme: Do you have any problems with x64 installer?

src/app/windows.utils.ts
```diff
@@ -19,18 +26,15 @@ export function getWindowsRegistryItem(hive: HKEY, path: string, name: string):
 		throw new Error('getWindowsRegistryItem is only available on Windows')
 	}
 
-	if (process.arch !== 'x64') {
-		throw new Error('getWindowsRegistryItem is only available on x64')
-	}
```
ShGKme, line 24: Let's add arm64 support and keep this check, but check for both: x64 and arm64 as allowed archs.

src/app/windows.utils.ts
```diff
@@ -19,18 +26,15 @@ export function getWindowsRegistryItem(hive: HKEY, path: string, name: string):
...
 		const require = createRequire(import.meta.url)
-		const windowRegistry = require('@vscode/windows-registry/prebuilds/win32-x64/@vscode+windows-registry.node')
+		const windowRegistry = require(`@vscode/windows-registry/prebuilds/win32-${process.arch}/@vscode+windows-registry.node`)
 
 		return windowRegistry.GetStringRegKey(hive, path, name)
 	} catch (error) {
 		// 'Unable to open registry key' is expected when the registry key does not exist
-		if ((error as Error).message !== 'Unable to open registry key') {
+		// 'MODULE_NOT_FOUND' is expected on architectures without a native prebuild (e.g., arm64)
+		if ((error as Error).message !== 'Unable to open registry key' && (error as NodeJS.ErrnoException).code !== 'MODULE_NOT_FOUND') {
```
ShGKme: Let's split package loading and Windows Registry reading into two try blocks.

Slightly longer, but more reliable catch slitting (since JS doesn't have a proper catch with predictable exception classes).

Something like:

```js
	let windowRegistry
	try {
		const require = createRequire(import.meta.url)
		windowRegistry = require(`@vscode/windows-registry/prebuilds/win32-${process.arch}/@vscode+windows-registry.node`)
	} catch {
  	console.error('Error on @vscode/windows-registry loading')
	  return undefined
	}

  try {
		return windowRegistry.GetStringRegKey(hive, path, name)
	} /* ... */
```

src/app/windows.utils.ts
```diff
@@ -10,6 +10,13 @@ export type HKEY = 'HKEY_CURRENT_USER' | 'HKEY_LOCAL_MACHINE' | 'HKEY_CLASSES_RO
 /**
  * Read Windows Registry item (only string values are supported)
  *
+ * Resolves the native `@vscode/windows-registry` prebuild for the current CPU architecture
+ * (e.g., `prebuilds/win32-x64/...`, `prebuilds/win32-arm64/...`).
+ * If no prebuild is shipped for the current architecture (or it cannot be loaded for any
+ * other reason), this fails gracefully and returns `undefined` instead of throwing -
+ * this function backs an optional feature (managed configuration via Registry policy)
+ * and must never crash the app on startup because of a missing native module.
+ *
```
ShGKme: Resolving `@vscode/windows-registry` is not the purpose of the function. It reads the value from the Windows Registry. All the other comments here are implementation details and comments for developers.

Let's keep it clean with `Read Windows Registry item`.

ShGKme (commented):
Thank you and sorry we missed the arm64. One day we'll have a Windows arm64 device for testing =D

Instead of ignoring arm64, let's actually add support for it.
I'm preparing a new release of the package with arm64 support.

You can test it locally with:
```sh
npm i github:nextcloud-deps/vscode-windows-registry#fix/build-arm64
```

If you have any errors on install, just ignore scripts:
```sh
npm i --ignore-scripts github:nextcloud-deps/vscode-windows-registry#fix/build-arm64 
```

ShGKme:
Asking just in case. If this PR is AI Assisted, please, add `Assisted-by: AGENT_NAME:MODEL_VERSION` to the commit description.

Details: https://github.com/nextcloud/server/blob/master/.github/CONTRIBUTING.md#ai-assisted-contributions

forge.config.js
aequivalent2: https://github.com/electron-userland/electron-wix-msi/issues/150

This issue states that, while not officially supported, it does work.

aequivalent2:
Ive added a few things

aequivalent2:
Is there anything what has to be done now?

aequivalent2:
Sorry is @ShGKme is there a problem with my pull request?

ShGKme:
@aequivalent2 To have a complete arm64 support here, we needed a new release of the `nextcloud-deps/vscode-windows-registry` and to have it updated in the app.

It was only done at the end of the last week in:
- https://github.com/nextcloud/talk-desktop/pull/1822

Now you can rebase your PR onto `main`, and ar64 support you added should actually work.

src/app/windows.utils.ts
```diff
@@ -10,6 +10,10 @@ export type HKEY = 'HKEY_CURRENT_USER' | 'HKEY_LOCAL_MACHINE' | 'HKEY_CLASSES_RO
 /**
  * Read Windows Registry item (only string values are supported)
  *
+ * Returns `undefined` instead of throwing if the value cannot be read - this function backs
+ * an optional feature (managed configuration via Registry policy) and must never crash the
+ * app on startup.
+ *
```
ShGKme: This comment is unnecessary for the function documentation.

"Returns `undefined` instead of throwing`" would be important if it were expected to throw.

And the rest (must never crash the app on startup) is the current bug description, not a property of the function (this function doesn't know it has anything to do with the app startup).


```suggestion
```

src/app/windows.utils.ts
```diff
@@ -19,14 +23,24 @@ export function getWindowsRegistryItem(hive: HKEY, path: string, name: string):
...
-	if (process.arch !== 'x64') {
-		throw new Error('getWindowsRegistryItem is only available on x64')
+	if (process.arch !== 'x64' && process.arch !== 'arm64') {
+		throw new Error('getWindowsRegistryItem is only available on x64 and arm64')
 	}
 
+	// The native `@vscode/windows-registry` prebuild is resolved per CPU architecture
+	// (e.g., `prebuilds/win32-x64/...`, `prebuilds/win32-arm64/...`). Loading it is kept in its
+	// own try/catch so a missing prebuild for the current architecture can never be confused
+	// with a registry-read error below.
```
ShGKme: Same here. The part about `process.arch` is visible directly from the code below, and the rest duplicates the PR's review comments rather than an edge case or explanation of an unclear code.


```suggestion
```

forge.config.js
```diff
@@ -244,7 +244,10 @@ module.exports = {
 			manufacturer: BUILD_CONFIG.companyName,
 			shortName: BUILD_CONFIG.applicationNameSanitized,
 			upgradeCode: BUILD_CONFIG.winUpgradeCode,
-			arch: 'x64', // electron-wix-msi defaults to x86 and arm64 is not supported
+			// electron-wix-msi (>=5.1.3, in use here) supports 'x86' | 'x64' | 'arm64' as Platform
+			// via WiX Toolset >=3.14's candle.exe. Use the actual packaging target instead of a
+			// hardcoded 'x64', otherwise win32-arm64 builds ship an MSI mislabeled as x64.
```
ShGKme: As was said, this comment is misleading. `'x86' | 'x64' | 'arm64'` is not the correct type for the `arch` option. According to its typing, the package documentation and the source code, it expects `arch` to be either `x86` or `x64`. There was a comment from a user in an issue that it works – it is not a contract, and not even a packager's contributor statement.

If it actually works (please, also test manually), it should be described in the comment as it is.

For example:

```suggestion
			// electron-wix-msi expects only 'x86' | 'x64' here.
			// However, it uses WiX Toolset >=3.14 which also supports `arm64`.
			// Passing `arm64` does not break packaging despite no explicit support from the electron-wix-msi
```

ShGKme (commented):
Could you adjust some comments to not have a misleading description for the `electron-wix-msi` and to not have too many extra comments? Comments describing the code are an additional part of the code that needs to be maintained (be consistent with the code changes in the future).

All the rest is good, but a rebase onto main is needed to test with the actual `@vscode/windows-registry` version, supporting `arm64`.

aequivalent2:
Hello, the new electron wixx 8.0.0 Beta added an Arch Check which prevents arm64, thats why i removed this change.
Further More i have deleted unecessary  comments

ShGKme (approved):
Thanks!

ShGKme:
> the new electron wixx 8.0.0 Beta added an Arch Check which prevents arm64, thats why i removed this change.

~~Hmm, [who](https://github.com/electron/forge/blame/v8.0.0-alpha.10/packages/maker/wix/src/MakerWix.ts) could have done that~~

image

I may propose to add the `arm64` here, but it might be complicated without having `arm64` explicitly supported by the `electron-wix-msi`. Even if it works, it must be a part of the contract.

As far as I understand, having an x64 installer should have no negative effect on the installed arm64 application. The app itself is `arm64`, it is inthalled in the same `C:\Program Files`. So the only problem is the installer name (it is generated as `Windows-x64-`) and, maybe, sharing the upgrade code.

Could you confirm, there are no issues with the x64 installer for the arm64 app on Windows ARM on your side?

aequivalent2:
With this check it is no longer possible to package the msi, you can see the error in the workflow run below

https://github.com/aequivalent2/nextcloud-desktop-arm64/actions/runs/30529848769/job/90829330314

ShGKme:
I understand passing `arm64` won't work here (as per the link above, there is an explicit check for only the supported values).

But it is unclear from the action if it's running with `arch: 'x64'`. 

> With this check it is no longer possible to package the msi, you can see the error in the workflow run below
> 
> https://github.com/aequivalent2/nextcloud-desktop-arm64/actions/runs/30529848769/job/90829330314

Are you sure, this is a run with `arch: 'x64'`? Not with your previous commit?

ShGKme:
Could you also remove the 2 last commits?
The 3rd just reverts the 2nd, so they change nothing together.

aequivalent2:
Hello, this run worked with the previous version without the check, now there is a failure

https://github.com/aequivalent2/nextcloud-desktop-arm64/actions/runs/30432968591/job/90514166207

package:arm64:msi was possible before but is not possible anymore i guess

aequivalent2:
I would like to add, that an solution for this problem would be nice, because msi builds are very usefull

aequivalent2:
> Could you also remove the 2 last commits? The 3rd just reverts the 2nd, so they change nothing together.

i have done it

aequivalent2:
What is the problem right now?

aequivalent2:
Is there anything left to do @ShGKme

ShGKme:
@aequivalent2 Sorry, from your comment with the action example I understood, that the build doesn't work anyway due to Wix MSI parameter. Did I get it wrong, and the current change is enough for you to make an arm64 build?

aequivalent2:
The exe is working the msi is not buildable, i thought there is a chance to Figuren out an solution

ShGKme (approved):

ShGKme:
I have rebased the branch onto actual main, no additional changes.

ShGKme:
> The exe is working the msi is not buildable, i thought there is a chance to Figuren out an solution

Since it works for the exe at least , I'm merging it, thank you.

For the `.msi` we have 2 options:
1. Wait for the change on the `electron-wix-msi` being reviewed/fixed/merged and then we can ask the electron-forge team to allow this option again
   - https://github.com/electron-userland/electron-wix-msi/pull/197
   - The last release was 4 years ago. A bit unlikely... I'll try to propose the change on the forge side even if the library doesn't officially support it
2. Stop using both libraries, building msi from scratch.
   - Since we plan to switch to `.msix` and WiX v3 ios EOL anyway, the effort doesn't make much sense to me.

aequivalent2:
Thank you for your support, if i can assist with anything let me know : )

PR #1821, opened 2026-07-22
build(deps-dev): override icon-gen's sharp to ^0.35.0

ShGKme:
### ☑️ Resolves

- `icon-gen` has `^0.33.4`
- With `0.y.z` version `^` only updates patches, not minors like with `^1.0.0` and above
- There were no braking changes in these minor versions
- Fixes 1 high audit report warning


Antreesy (approved):
Tested with Win

PR #1822, opened 2026-07-23
build(deps): nextcloud-deps/vscode-windows-registry v1.2.0-prebuild.3

ShGKme:
- Updates downgraded version
- Fixes arm64 support
- Fixes optional install due to broken `install` script

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "zx": "^8.8.5"
   },
   "optionalDependencies": {
-    "@vscode/windows-registry": "github:nextcloud-deps/vscode-windows-registry#v1.2.0-prebuild"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -110,7 +110,7 @@
     "zx": "^8.8.5"
   },
   "optionalDependencies": {
-    "@vscode/windows-registry": "github:nextcloud-deps/vscode-windows-registry#v1.2.0-prebuild"
```
Antreesy, line 113: Have you found a way to prevent downgrading?

Antreesy (approved):

package.json
ShGKme: I hope that it was due to:
- Broken install script
- Not updating a version in package.json

ShGKme: If it is still broken and dependabot downgrades again, I'll add to the ignore

PR #1857, opened 2026-08-12
Save and restore main Talk window position and size

### LINKED ISSUE #15: Save and restore window position and size on reopen

xhon-pelushi:
## Summary
- Persist the main Talk window bounds in app config while moving/resizing/closing.
- Restore saved bounds on reopen when they are valid and still visible on a display.
- Fall back to the default size when saved bounds are missing or off-screen.

Fixes #15

## Test plan
- [ ] Open Talk Desktop, move and resize the window, quit, and reopen — position/size should restore.
- [ ] Disconnect/move off a secondary display and reopen — should not open off-screen.
- [ ] Fresh config without saved bounds still opens at the default size.

### DIFF
--- a/src/app/AppConfig.ts
+++ b/src/app/AppConfig.ts
@@ -11,6 +11,13 @@ import { isMac } from './system.utils.ts'
 
 const APP_CONFIG_FILE_NAME = 'config.json'
 
+type WindowBounds = {
+	x: number
+	y: number
+	width: number
+	height: number
+}
+
 // Memoize the path to the application config file
 let APP_CONFIG_FILE_PATH: string | null = null
 
@@ -40,6 +47,11 @@ export type AppConfig = {
 	 * Last application version when the app was running
 	 */
 	lastAppVersion?: string
+
+	/**
+	 * Last position and size of the main Talk window.
+	 */
+	talkWindowBounds?: WindowBounds
 
 	/**
 	 * List of the accounts in the app in the {user}@{server} or {server} format,
--- a/src/talk/talk.window.js
+++ b/src/talk/talk.window.js
@@ -3,9 +3,9 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-const { BrowserWindow } = require('electron')
+const { BrowserWindow, screen } = require('electron')
 const { setupTray } = require('../app/app.tray.js')
-const { getAppConfig } = require('../app/AppConfig.ts')
+const { getAppConfig, setAppConfig } = require('../app/AppConfig.ts')
 const { applyContextMenu } = require('../app/applyContextMenu.js')
 const { applyDownloadHandler } = require('../app/downloads.ts')
 const { applyExternalLinkHandler } = require('../app/externalLinkHandlers.ts')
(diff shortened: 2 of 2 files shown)


src/talk/talk.window.js
```diff
@@ -15,18 +15,115 @@ const { TITLE_BAR_HEIGHT } = require('../constants.js')
 const { BUILD_CONFIG } = require('../shared/build.config.ts')
 const { getBrowserWindowIcon } = require('../shared/icons.utils.js')
 
+const SAVE_WINDOW_BOUNDS_DELAY = 500
```
ShGKme, line 18: Thanks for adding the debounce!

Let's make it even slower to reduce the number of disk writes.
For example, 15_000 or even more.

They delay here is the time between the last window move and the app being closed. A short delay only helps when a user moves the window, then within half a second closes the app, and reopens it, expecting the position from the right before quit.

src/talk/talk.window.js
```diff
@@ -15,18 +15,115 @@ const { TITLE_BAR_HEIGHT } = require('../constants.js')
...
+	}
+
+	const savedBounds = {
+		...bounds,
+		width: Math.max(bounds.width, minimumSize.minWidth),
+		height: Math.max(bounds.height, minimumSize.minHeight),
+	}
+
+	return isVisibleOnAnyDisplay(savedBounds) ? savedBounds : undefined
+}
```
ShGKme, line 71: It searches for the first window it can fit, and if nothing is found, it fallbacks to the default size.

This means, it may always choose the wrong display just because the wrong one was first and it fits.

Let's adjust the algorithm here a bit:

1. Check the exact display that was used the last time (it should be saved as well)
2. If there is no saved display (or it is not connected), use the **primary** display
3. If it doesn't fit anymore (for example, the user changed the resolution or scaling) - try to fit instead of choosing another display. For example, if the window was 2560 px wide and now the display is just 1920px, instead of using another display, we can decrease the width to 1920px.

If `isMaximized` or `isFullscreen` was set, the size doesn't matter.

src/app/AppConfig.ts
```diff
@@ -11,6 +11,13 @@ import { isMac } from './system.utils.ts'
 
 const APP_CONFIG_FILE_NAME = 'config.json'
 
+type WindowBounds = {
+	x: number
+	y: number
+	width: number
+	height: number
+}
```
ShGKme, line 19: Let's add even more properties here to restore the state.
1. `isMaximized` - to restore the window maximized if it was maximized
2. `isFullScreen` - for macOS as another "maximized" state
3. `displayId` - to save the exact display the Talk window was open on

ShGKme (changes requested):
Hi, thank you for your contribution. This works and something we definitely want to have. But the current implementation misses some state properties and may miss the correct display on multiple displays setup. Could you adjust the solution?

ShGKme:
I have just found out, that 10 days ago Electron released v44 with 

> - Added a Save/Restore Window State API for cross-platform use. [#52270](https://github.com/electron/electron/pull/52270)

https://releases.electronjs.org/release/v44.0.0

So, Electron has it as a built-in feature now.

We don't use Electron 44 due to a bug on Linux. But it seems it was also fixed 2 weeks ago in 43.5.1 and the fix is included into Electron 44. So, we can upgrade now, and implement it via the built-in feature.

ShGKme:
We have upgraded Electron. Now built-in `windowStatePersistence` is available.

ShGKme:
Hi @xhon-pelushi,
It's been a while, and since the feature is now included into Electron, we will go with the built-in solution and close this PR. Thanks for your contribution and sorry for a long review.

PR #1862, opened 2026-08-18
fix: include webp as 'asset/inline' for webpack

Antreesy:
### ☑️ Resolves

- images were added as a new asset type with dashboard redesign

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image


ShGKme (approved):
Thanks

Issue #1156, opened 2025-03-06
Splash Screen Stuck


apesorguk:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Steps to reproduce
1. Instal latest version of Talk on windows.
2. Start Nextcloud Talk
3. Stays on splash and never progresses. 

### Expected behaviour
Load into Nextcloud Talk Client

### Actual behaviour

![Image](https://github.com/user-attachments/assets/bb64bc65-7c64-4780-b36d-e9aafdcdf53f)

## Diagnosis and logs

Can not produce any as it will not load.

ShGKme:
@apesorguk Could you click on the splash screen, then press `CTRL+SHIFT+I`, and then on the following buttons?

![Image](https://github.com/user-attachments/assets/e44e89d1-6b54-4474-8638-47dfd321bde8)

This will open dev tools in a new window. Do you have any errors in the `console` Tab?

Omaha2002:
I have the same issue on Macos:

<img width="1425" alt="Image" src="https://github.com/user-attachments/assets/2b37a8e4-76af-473f-abdd-7f2b3ee6fcc0" />

Omaha2002:
For Macos delete: ~/Library/Application Support/Nextcloud Talk directory

Have to fill in the server again afterwards

Antreesy:
> So I can confirm the issue is not just isolated to Windows but also my Linux machines. So for the moment, we are just using talk in the browser and phone as it's a pain to get Nextcloud Talk App working on laptops and PCs. 👎

Could you give a try to instructions from https://github.com/nextcloud/talk-desktop/issues/1156#issuecomment-2704539879 again, so we can get more information on what's happening on your end? `CTRL+SHIFT+I` combination works for me on stable build

ShGKme:
> I did that, and in the console, there are no logs or errors.

Try to reload after opening.

If there is no error, please, switch to the Network tab and check for a request that got stuck.

bakiburakogun:
[...]

PR #1883, opened 2026-09-02
fix(window): apply the ready-to-show fallback on all platforms

### LINKED ISSUE #1882: Windows: app stays on the splash screen after updating to 2.2.x, 'ready-to-show' never fires
## Steps to reproduce
1. Talk Desktop v2.1.1 is installed on Windows.
2. Update to v2.2.4.
3. Start the application.

### Expected behaviour
The splash screen closes and the main (Talk) window is shown.

### Actual behaviour
The red splash screen stays forever. Nothing else appears, the app has to be killed from the task manager.

We ship a branded build of talk-desktop (Kırlangıç) for our organisation, so we could look into it a bit:

- The splash renderer does its job and the main window is created, but the main window's `ready-to-show` event never fires, so `welcomeWindow.close()` / `mainWindow.show()` in `main.js` are never reached.
- It looks like the same class of problem as #1710 / electron/electron#48859 (`ready-to-show` not firing when the GPU process fails to initialise), just on Windows instead of Wayland. v2.2.4 moved from Electron 38 to Electron 43, which changed the Windows GPU path. Running the Windows binary under Wine, where the GPU process exits right after start, the app stays on the splash screen too, and with the fallback patched in, the log shows `ready-to-show` never firing for the welcome and main windows.
- `onReadyToShow()` in `src/app/utils.ts` only applies the `did-finish-load` fallback when `isWayland` is true, so on Windows there is no fallback at all.

Applying the fallback on every platform (with a hard timeout as a last resort) fixes it for us: with that change the same machine starts normally. See #1883.

## Diagnosis and logs

A diagnosis report cannot be produced because the app never gets past the splash screen. Client: build based on v2.2.4, Windows x64.

bakiburakogun:
### ☑️ Resolves

- Fix: #1882
- Possibly related: #1156

`ready-to-show` is not only unreliable on Wayland. We hit exactly the same symptom on Windows after moving our branded build from v2.1.1 to v2.2.4: the splash screen stays forever because the main window never gets `ready-to-show`. v2.2.4 comes with Electron 43; under Wine, where the GPU process exits right after start, the same thing happens and the added log line confirms `ready-to-show` never fires, which points to the same GPU-related cause as electron/electron#48859.

This applies the existing `did-finish-load` fallback on all platforms instead of Wayland only, and adds a 15 s hard timeout so a window can never stay hidden forever. On platforms where `ready-to-show` works nothing changes: it still wins the race, the timers are cleared. On Wayland the behaviour is the same as before (fallback runs right after `did-finish-load`); elsewhere the fallback waits 1.5 s after load to avoid showing an unpainted window in the normal case.

`did-fail-load` is treated as "show now" as well, so a failed load ends up visible instead of an invisible window.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Splash screen stays forever on affected Windows machines | Main window is shown ~1.5 s after the page has loaded

### DIFF
--- a/src/app/utils.ts
+++ b/src/app/utils.ts
@@ -10,7 +10,6 @@ import { APP_ORIGIN } from '../constants.js'
 import { BUILD_CONFIG } from '../shared/build.config.ts'
 import { getAppConfig } from './AppConfig.ts'
 import { appData } from './AppData.js'
-import { isWayland } from './system.utils.ts'
 
 /**
  * Get the scaled window size based on the current zoom factor
@@ -179,16 +178,29 @@ export function getTitleBarSymbolColor(backgroundColor: string = BUILD_CONFIG.ba
  * @param callback - Callback
  */
 export function onReadyToShow(window: BrowserWindow, callback: () => void) {
-	if (!isWayland) {
-		window.once('ready-to-show', callback)
-		return
-	}
-
-	// 'ready-to-show' may not be triggered on Wayland due to GPU sync issues in Chromium on some GPUs and VMware
-	// A workaround: wait for the page load and then for a tick
+	// 'ready-to-show' may never be emitted when the GPU process fails to initialize
+	// (seen with some GPUs and VMware on Wayland, but not tied to the platform as such),
+	// which would leave the window hidden forever.
+	// Race it against "page loaded + a tick", which is late enough not to show a blank window.
 	// See: https://github.com/electron/electron/issues/48859
-	window.webContents.once('did-finish-load', async () => {
-		await window.webContents.executeJavaScript('new Promise((resolve) => setTimeout(resolve, 0))')
+	const readyToShow = new Promise<string>((resolve) => {
+		window.once('ready-to-show', () => resolve('ready-to-show'))
+	})
+	const pageLoaded = new Promise<string>((resolve) => {
+		window.webContents.once('did-finish-load', () => {
+			window.webContents.executeJavaScript('new Promise((resolve) => setTimeout(resolve, 0))')
+				.catch(() => {}) // The window may be gone already
+				.then(() => resolve('did-finish-load'))
+		})
+	})
+
+	Promise.race([readyToShow, pageLoaded]).then((winner) => {
+		if (window.isDestroyed()) {
+			return
+		}
+		if (winner !== 'ready-to-show') {
+			console.warn('[window] "ready-to-show" was not emitted, showing the window after the page load')
+		}
 		callback()
 	})
 }


src/app/utils.ts
```diff
@@ -179,16 +179,32 @@ export function getTitleBarSymbolColor(backgroundColor: string = BUILD_CONFIG.ba
...
+	window.once('ready-to-show', () => run('ready-to-show'))
+	window.webContents.once('did-finish-load', () => {
+		// On Wayland 'ready-to-show' is known to be unreliable - show right after the load (as before)
+		// Elsewhere give 'ready-to-show' a moment to avoid showing a not-yet-painted window in the normal case
+		timers.push(setTimeout(() => run('did-finish-load'), isWayland ? 0 : 1500))
+	})
+	window.webContents.once('did-fail-load', (event, errorCode, errorDescription) => {
+		run(`did-fail-load ${errorCode} ${errorDescription}`)
 	})
+	timers.push(setTimeout(() => run('timeout'), 15_000))
```
ShGKme: New workaround doesn't seem clear to me.
- Where do `1500` and `15_000` constants come from?
- Was there a problem with the original wait for the page load and then for a single tick instead of waiting for 1.5 seconds?
- Why is a window shown anyway after timeout? If there is a need to wait for 15 seconds, then the workarounds didn't work. In this case there is a need for another workaround.
- Why is `did-fail-load` a reason to show the window?

The race implementation on timers can be simplified with built-in `Promise.race`, aka:
```ts
Promise.race([
	new Promise((resolve) => window.once('ready-to-show', resolve)),

	new Promise((resolve) => window.webContents.once('did-finish-load', async () => {
		await window.webContents.executeJavaScript('new Promise((resolve) => setTimeout(resolve, 0))')
		resolve()
	})),
])
```

ShGKme (changes requested):
Thank you for investigation the issue. The Electron issue is clear. But the workaround needs some adjustments and clarifications.

bakiburakogun:
Fair points, thanks for the review. The constants were guesses rather than measurements, so I dropped them and went with your race as is, on all platforms. The 15 s timeout is gone too: that logic belongs to a startup watchdog in our branded build, not here, and if nothing loads within 15 seconds an empty window doesn't help anyone. Same for `did-fail-load`: Chromium paints the error page anyway, so `ready-to-show` fires.

One thing I should be upfront about: the Windows report that started this turned out to be caused by our own updater (see #1882), so I don't have a confirmed non-Wayland case anymore. What's left is that the GPU init failure behind electron/electron#48859 isn't tied to Wayland as such, and the race costs nothing when `ready-to-show` comes first, which in our logs is about 2 ms after `did-finish-load`. If you'd rather keep the workaround Wayland-only, I'm fine with closing this.

ShGKme (approved):
Looks good.

The issue only seems to happen on Wayland, but a general workaround with a race seems better. For Wayland users, it allows using the original `ready-to-show` event when it works without an unneeded workaround. For other users it is fail-safe in case this happens with other systems.

PR #1884, opened 2026-09-02
fix(authentication): login fails with UTF-8 characters in username from an external identity provider

### LINKED ISSUE #1360: Login fails in Talk Desktop App when username contains umlauts (ä, ö, ü) – 401 Unauthorized
### Issue Summary

When logging into the **Nextcloud Talk Desktop App** (Windows), users with non-ASCII characters (e.g. German umlauts like `ä`, `ö`, `ü`) in their **username** experience the following error:

> "Login successful, but something went wrong."

The same users **can log in successfully via the browser** and use Talk without issue.

---

### Steps to Reproduce

1. Use a Nextcloud server with LDAP authentication
2. Ensure a user has a login name like `müller` (with umlaut)
3. Attempt login with the **Talk Desktop App** (tested: Windows, v1.1.7)
4. The app reports successful login, but immediately fails with error above
5. In the server logs (`access.log`), the username appears misencoded:

`\xdc` is the Latin-1 encoding of `Ü`, which suggests **incorrect character encoding** by the client.

---

### Expected Behavior

- The app should handle usernames with UTF-8 characters correctly
- Users with valid credentials should be able to log in regardless of non-ASCII characters in usernames

---

### Actual Behavior

- Users with umlauts in their usernames are rejected with HTTP 401
- Only the Desktop Talk app is affected (browser interface works)

---

### Environment

- Nextcloud Server: [Insert your version here]
- Talk App: Desktop v1.1.7 (Windows)
- LDAP backend with usernames containing UTF-8 characters
- Working setup otherwise (SPF, DKIM, LDAP, browser login OK)

---

### Suggested Fix

Ensure the Talk desktop app:
- Properly encodes login credentials (especially the username) in UTF-8
- Sends compliant WebSocket and HTTP headers for OCS calls with UTF-8 values

---

### Related

- Possibly related to: issues with international characters or encoding bugs in Electron apps

bakiburakogun:
### ☑️ Resolves

- Fix: #1360

`webRequestInterceptor.js` builds the `Authorization: Basic …` header with `btoa()`, which only handles Latin-1. For a login name with characters outside of Latin-1 (Turkish `ş`, `ı`, `ç`, Cyrillic, …) it throws `InvalidCharacterError`, and for Latin-1 characters such as `ü`/`ö` it produces Latin-1 bytes while the server decodes the header as UTF-8. Either way the first authenticated request after the login flow fails with 401, which is exactly the "Login successful, but something went wrong" reported in #1360 (LDAP login names with umlauts).

This encodes the credentials with `Buffer.from(..., 'utf8')` instead. Quick check in Node:

```
btoa('müller:pw')            -> "bfxsbGVyOnB3"  (decodes to "m\xFCller", not UTF-8)
btoa('şükrü:pw')             -> InvalidCharacterError
Buffer.from('şükrü:pw', 'utf8').toString('base64') -> works, server reads "şükrü"
```

No behaviour change for ASCII user names (same output as before).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Login with `müller` / `şükrü` ends in 401 after the login flow | Login works

### DIFF
--- a/src/app/webRequestInterceptor.js
+++ b/src/app/webRequestInterceptor.js
@@ -27,7 +27,9 @@ function enableWebRequestInterceptor(serverUrl, { credentials } = {}) {
 				requestHeaders: {
 					...details.requestHeaders,
 					Origin: new URL(serverUrl).origin,
-					Authorization: `Basic ${btoa(`${credentials.user}:${credentials.password}`)}`,
+					// btoa() only handles Latin-1: it throws on characters like ş or ı and encodes ü/ö as Latin-1,
+					// while the server decodes the header as UTF-8 - see #1360
+					Authorization: `Basic ${Buffer.from(`${credentials.user}:${credentials.password}`, 'utf8').toString('base64')}`,
 					'OCS-APIRequest': 'true',
 				},
 			})


ShGKme (approved):
Thanks

PR #1897, opened 2026-09-07
fix(title-bar): adjust menus with the new nextcloud-vue popover design

ShGKme:
### ☑️ Resolves

- Bump `nextcloud-vue`
  - Applies new design to popovers
- The position was slightly off
  - The title bar is `4px` larger than the default `--default-clickable-area + --default-grid-baseline`
  - `UserMenu` can be adjusted with the `distance` prop
  - `MainMenu` was adjusted with a hacky solution by increasing the button size
- Alternatives:
  - Mount the menu inside the title bar and override the position with `:deep()`
  - Decrease the title bar height but this is too visible. Will be done during larger title bar redesign

### 🖼️ Screenshots

Before | Without the hack | After
-- | -- | --
image | image | image


src/talk/renderer/TitleBar/components/UserMenu.vue
```diff
@@ -74,7 +74,8 @@ function handleUserStatusChange(status: UserStatusStatusType) {
 			:container="userMenuContainer"
 			:popperHideTriggers="(triggers: string[]) => [...triggers, 'click']"
 			:triggers="[]"
-			noAutoFocus>
+			noAutoFocus
+			:distance="4 /* default distance */ + 4 /* additional title bar buffer */">
```
Antreesy: Maybe move it out of template?

Antreesy (approved):

src/talk/renderer/TitleBar/components/UserMenu.vue
ShGKme: I kept it in the template intentionally, because it has value only in this context: it is not reused, not a complicated computation, and describes only this specific prop in this specific component.

Where would you move it?

Antreesy: just as const to script setup. I personally find it cumbersome to see comments in this line

ShGKme: Done

Antreesy (approved):

PR #1898, opened 2026-09-08
build(deps-dev): bump webpack 1.110.1 -> 1.110.3

ShGKme:
### ☑️ Resolves

- Fix: production build fails with `ERROR: "minify" must be a boolean`
- Regression from: https://github.com/nextcloud/talk-desktop/pull/1895
  - https://github.com/webpack/webpack/issues/21879
- Fixed in Webpack 1.110.3


Antreesy (approved):
Confirmed

PR #1901, opened 2026-09-10
feat(Viewer): allow to paginate Viewer media list

Antreesy:
### ☑️ Resolves

- list is already provided by chat messages and shared items tab, just wasn't used by client
- loadMore normalization is temp until fixed in spreed side

Assisted-by: ClaudeCode:claude-sonnet-5

### 🖼️ Screenshots

🏡 After

https://github.com/user-attachments/assets/57030f41-2344-4767-80eb-1d1bac336a8c

TODO

loadMore is broken on `spreed`, need to normalize there


src/talk/renderer/Viewer/ViewerApp.vue
```diff
@@ -63,7 +140,11 @@ defineExpose({
 		size="full"
 		closeButtonOutside
 		dark
-		@close="close">
+		:hasPrevious="hasPrevious"
+		:hasNext="hasNext"
```
ShGKme:
```suggestion
		:hasPrevious
		:hasNext
```

src/talk/renderer/Viewer/ViewerApp.vue
```diff
@@ -42,10 +55,74 @@ function open(options = {}) {
...
+	}
+}
+
+/**
+ * Normalize a file object returned by `loadMore` to the shape the Viewer expects.
+ * Talk's `useViewer` composable normalizes the initial `list`/`fileInfo` (raw `id`/`path`/`name`/`mimetype`
+ * to `fileid`/`filename`/`basename`/`mime`) but forwards `loadMore` unwrapped, so items it resolves with
+ * still use the raw field names and need the same normalization applied here.
+ *
+ * FIXME should be normalized by spreed
```
ShGKme: From the description it's unclear why this is not normalized by spreed and why it works in the original Viewer

ShGKme (commented):
I guess, load more is used to go through the history? But only one direction?

src/talk/renderer/Viewer/ViewerApp.vue
ShGKme: Got it

Antreesy:
> I guess, load more is used to go through the history? But only one direction?

yes, endpoint has only support one direction - get older from given lastMessageId

ShGKme (approved):

